PR #24, opened 2023-07-21
fix save full state dict checkpoint

jeonsworld:
- Resolved an issue with saving checkpoints as **FULL_STATE_DICT** type in FSDP settings.

- `cfg.checkpoint_dir` does not exist, so Change `cfg.checkpoint_dir` to `cfg.dist_checkpoint_root_folder` and `cfg.dist_checkpoint_folder`

- Since the `verbose` variable does not exist, errors occur in some sections, so add the `verbose` variable to training_config

- import optimize


jeonsworld:
CLA

HamidShojanazeri (approved):
Thanks @jeonsworld for the PR.

PR #39, opened 2023-07-25
Fsdp inference checkpoints

HamidShojanazeri:
Adding converter script for converting FSDP sharded state dict to HF model for inference.


inference/inference.py
```diff
@@ -12,8 +12,17 @@
...
+from model_utils import load_model, load_peft_model, load_llama_from_config
+from accelerate import init_empty_weights
+# Get the current file's directory
+current_directory = os.path.dirname(os.path.abspath(__file__))
 
+# Get the parent directory
+parent_directory = os.path.dirname(current_directory)
+
+# Append the parent directory to sys.path
+sys.path.append(parent_directory)
```
chauhang, line 24: Please move the code for path setting inside main / function

inference/inference.py
```diff
@@ -83,7 +92,7 @@ def main(
     if peft_model:
         model = load_peft_model(model, peft_model)
 
-    model.eval()
+    # model.eval()
```
chauhang, line 95: What is the reason for commenting this?

inference/inference.py
```diff
@@ -12,8 +12,17 @@
...
+from accelerate import init_empty_weights
+# Get the current file's directory
+current_directory = os.path.dirname(os.path.abspath(__file__))
 
+# Get the parent directory
+parent_directory = os.path.dirname(current_directory)
+
+# Append the parent directory to sys.path
+sys.path.append(parent_directory)
+from model_checkpointing import load_sharded_model_single_gpu
```
chauhang, line 25: Where is load_sharded_model_single_gpu getting used in the inference.py?

docs/inference.md
```diff
@@ -34,6 +34,15 @@ The inference folder also includes a chat completion example, that adds built-in
 python chat_completion.py --model_name "PATH/TO/MODEL/7B/" --prompt_file chats.json  --quantization --use_auditnlg
 
 ```
+## Loading back FSDP checkpoints
+
+In case you have fine-tuned your model with pure FSDP and saved the checkpoints with "SHARDED_STATE_DICT" as shown [here](../configs/fsdp.py), you can use this converter script to convert the FSDP Sharded checkpoints into HuggingFace checkpoints. This enables you to use the inference script normally as mentioned above.
+**To convert the checkpoint use the following command**:
+```bash
+ python checkpoint_converter_fsdp_hf.py --model_name  PATH/to/FSDP/Checkpoints --save_dir PATH/to/save/checkpoints --model_path PATH/or/HF/model_name
```
chauhang, line 42: checkpoint_converter_fsdp_hf.py file is not there in this PR

chauhang (changes requested):
@HamidShojanazeri Thanks for getting this started. Please see comments inline

inference/inference.py
HamidShojanazeri: just removed was legacy code.

docs/inference.md
chauhang, line 42: Will be better to rename the parameters to reflect their usage for conversion. In this case original FSDP checkpoints already exists. Suggest changing to something like below: 

1st param -- fsdp_checkpoint_root PATH/to/FSDP/Checkpoints
2nd param --fsdp_model_name  PATH/to/saved/checkpoints
3rd param --consolidated_model_path Destination location for saving the consolidated model in HF format

inference/checkpoint_converter_fsdp_hf.py
```diff
@@ -0,0 +1,44 @@
...
+# Append the parent directory to sys.path
+sys.path.append(parent_directory)
+from model_checkpointing import load_sharded_model_single_gpu
+
+def main(
+    model_name,
+    save_dir="", # Path to save the HF converted model checkpoints
+    model_path="" # Path/ name of the HF model that include config.json and tokenizer_config.json
+    ):
+    #load the HF model definition from config
```
chauhang, line 31: Who is generating this config file?

chauhang (changes requested):
Please add steps used for testing this out, including below along with output of your test for verification: 

1. command to run and generate the pure fsdp checkpoint
2. command for converting the checkpoint to HF format for inference
3. command for running the inference

inference/checkpoint_converter_fsdp_hf.py
HamidShojanazeri: This is basically the HF model-name/ checkpoints where they have the config.json

docs/inference.md
HamidShojanazeri: thanks Geeta, I think maybe we can leave 1&2 FSDP checkpoints as user may be confused how to organize/ recognize these. Changed the third to the suggested option.

HamidShojanazeri:
> Please add steps used for testing this out, including below along with output of your test for verification:
> 
> 1. command to run and generate the pure fsdp checkpoint
> 2. command for converting the checkpoint to HF format for inference
> 3. command for running the inference

done.

inference/checkpoint_converter_fsdp_hf.py
HamidShojanazeri: Added the train_param.yaml (saved during the finetuening) to by default search for model_name instead of asking user first.

docs/inference.md
```diff
@@ -34,6 +34,31 @@ The inference folder also includes a chat completion example, that adds built-in
...
+**To convert the checkpoint use the following command**:
+
+This is helpful if you have fine-tuned you model using FSDP only as follows:
+
+```bash
+torchrun --nnodes 1 --nproc_per_node 8  llama_finetuning.py --enable_fsdp --model_name /patht_of_model_folder/7B --dist_checkpoint_root_folder model_checkpoints --dist_checkpoint_folder fine-tuned --pure_bf16 
+```
+Then convert your FSDP checkpoint to HuggingFace checkpoints using:
+```bash
+ python checkpoint_converter_fsdp_hf.py --fsdp_checkpoint_path  PATH/to/FSDP/Checkpoints --consolidated_model_path PATH/to/save/checkpoints --HF_model_path_or_name PATH/or/HF/model_name
```
chauhang, line 49: Add inference folder name if running from the root directory  `python inference/check....

inference/checkpoint_converter_fsdp_hf.py
```diff
@@ -0,0 +1,63 @@
...
+        print(f"The file {train_params_path} does not exist.")
+        HF_model_path_or_name = input("Please enter the model name: ")
+        print(f"Model name: {HF_model_path_or_name}")
+    except Exception as e:
+        print(f"An error occurred: {e}")
+        
+        
+    #load the HF model definition from config
+    model_def = load_llama_from_config(HF_model_path_or_name)
+    print("model is loaded from config")
```
chauhang, line 52: nit - move the print before the actual loading (previous line) otherwise takes a long time to print the message...

chauhang (approved):

PR #44, opened 2023-07-25
adding active mem stat

HamidShojanazeri:
small PR to add active memory stat to the training loop. Also setting the BF16 default to false to avoid confusion.

### DIFF
--- a/configs/fsdp.py
+++ b/configs/fsdp.py
@@ -13,7 +13,7 @@ class fsdp_config:
     sharding_strategy: ShardingStrategy = ShardingStrategy.FULL_SHARD
     checkpoint_type: StateDictType = StateDictType.SHARDED_STATE_DICT  # alternatively can use SHARDED_STATE_DICT save one file per rank, and can resize the world-size.
     fsdp_activation_checkpointing: bool=True
-    pure_bf16: bool = True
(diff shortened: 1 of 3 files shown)


configs/fsdp.py
```diff
@@ -13,7 +13,7 @@ class fsdp_config:
     sharding_strategy: ShardingStrategy = ShardingStrategy.FULL_SHARD
     checkpoint_type: StateDictType = StateDictType.SHARDED_STATE_DICT  # alternatively can use SHARDED_STATE_DICT save one file per rank, and can resize the world-size.
     fsdp_activation_checkpointing: bool=True
-    pure_bf16: bool = True
+    pure_bf16: bool = False
```
chauhang, line 16: What is the reason for changing the default to False?

chauhang (commented):
@HamidShojanazeri Thanks for adding this support. Please see comments inline

configs/fsdp.py
HamidShojanazeri: I realized we are adding --pure_bf16 in our command, if we leave this to True and then user set --use_fp16 this one would be in effect by default as well, so it might be better to default to false and enable it as needed.

chauhang (approved):

PR #51, opened 2023-07-26
fixing the full state path in checkpoint handler+loss report calculation

HamidShojanazeri:
This PR fixes 
- the checkpoint path for full_state_dict and also removes the local_state_dict code
- loss report calculation, issue #68 .


utils/train_utils.py
```diff
@@ -174,7 +177,7 @@ def train(model, train_dataloader,eval_dataloader, tokenizer, optimizer, lr_sche
         results['avg_eval_prep'] = avg_eval_prep
         results['avg_eval_loss'] = avg_eval_loss
         
-
+    dist.barrier()
```
chauhang, line 180: Please check if this is still required, since the barrier has been added before and after the checkpoint already.

utils/train_utils.py
```diff
@@ -126,7 +125,7 @@ def train(model, train_dataloader,eval_dataloader, tokenizer, optimizer, lr_sche
         if train_config.run_validation:
             eval_ppl, eval_epoch_loss = evaluation(model, train_config, eval_dataloader, rank, tokenizer)   
             if train_config.save_model and eval_epoch_loss < best_val_loss:
-                
+                dist.barrier()
                 if  train_config.use_peft:
                     
                     print(f"we are in the saving the PEFT modules")
```
chauhang, line 131: For the PEFT model saving, does this need to happen on all the ranks or only the rank 0? Please check

HamidShojanazeri: as the model is wrapped in fsdp, this would not work, it hangs then after saving on rank0.

utils/train_utils.py
```diff
@@ -250,6 +267,7 @@ def setup_environ_flags(rank):
     os.environ["TORCH_SHOW_CPP_STACKTRACES"] = str(1)
     os.environ["NCCL_ASYNC_ERROR_HANDLING"] = str(1)
     os.environ["TORCH_DISTRIBUTED_DEBUG"] = "DETAIL"
+    os.environ['PYTORCH_CUDA_ALLOC_CONF']='expandable_segments:True'
```
chauhang, line 270: expandable_segments is only available for the PyTorch Nightlies at present. TORCH_DISTRIBUTED_DEBUG=DETAIL will be too verbose for regular training. Will be better to document these for additional troubleshooting tips

utils/train_utils.py
```diff
@@ -104,62 +100,97 @@ def train(model, train_dataloader,eval_dataloader, tokenizer, optimizer, lr_sche
...
             eval_ppl, eval_epoch_loss = evaluation(model, train_config, eval_dataloader, rank, tokenizer)   
             if train_config.save_model and eval_epoch_loss < best_val_loss:
-                
-                if  train_config.use_peft:
-                    
-                    print(f"we are in the saving the PEFT modules")
-                    model.save_pretrained(train_config.output_dir)   
-                    print(f"PEFT modules are saved in {train_config.output_dir} directory")
-                    
+                dist.barrier()
```
chauhang, line 140: Add check to do this for distributed multi-gpu case only

utils/train_utils.py
```diff
@@ -104,62 +100,97 @@ def train(model, train_dataloader,eval_dataloader, tokenizer, optimizer, lr_sche
...
                     if not train_config.use_peft and  train_config.save_optimizer:
                         model_checkpointing.save_optimizer_checkpoint(
-                            model, optimizer, rank, train_config, epoch=1
-                        )   
-                                
+                            model, optimizer, rank, train_config, epoch=epoch
+                        )
+                        print(" Saving the FSDP model checkpoints qnd optimizer using FULL_STATE_DICT")
+                        print("=====================================================")                     
+                dist.barrier()
```
chauhang, line 176: Add check to do this for distributed multi-gpu case only

chauhang (changes requested):
@HamidShojanazeri Thanks for the fixes. Please see comments inline.

utils/train_utils.py
HamidShojanazeri: Sure, I will add them to the FAQ and just comment them here instead of removing to leave some hints for user.

DhruvaBansal00:
Seems like this is supposed to fix #65? I am going to try this out and will keep you folks updated!

DhruvaBansal00:
Screenshot 2023-07-31 at 2 08 42 PM

Seeing the same issue with this PR too.

HamidShojanazeri:
@DhruvaBansal00 thanks for trying out the PR, I am not able to repro on my end can you pls share the command your running? and you env using`python -m "torch.utils.collect_env"`.

DhruvaBansal00:
@HamidShojanazeri I hadn't applied your changes correctly before. Just finished running three epochs and seems like everything is running as expected. Thanks for the changes and appreciate the fast turnaround :)

HamidShojanazeri:
> @HamidShojanazeri I hadn't applied your changes correctly before. Just finished running three epochs and seems like everything is running as expected. Thanks for the changes and appreciate the fast turnaround :)

Thanks very much @DhruvaBansal00 for confirmation.

chauhang (approved):
Thanks @HamidShojanazeri for all the fixes and updates. 

Thanks @jeonsworld your changes for [PR #24 ](https://github.com/facebookresearch/llama-recipes/pull/24) have also been rolled into this PR

PR #77, opened 2023-08-01
save cpu mem by leveraging FSDP rank0 broadcasting

lchu-ibm:
# What does this PR do?

for FSDP mode, this saves cpu memory by loading only one cpu copy of the model. This is specifically useful when using llama 70B as the current code would consume 2+ TB of cpu memory with 70B (70 * 4 * 8), which will cause cpu oom.

# Notes

1. This would require latest nightlies.  I vaguely remembered I hit various of issues with `sync_module_states`+`param_init_fn` in the past until the nightlies in the most recent months. 
2. ~~I wasn't sure what's the best in-general `_param_init_fn` we should use in the current version given the fast evolving PRs around meta device init.  maybe @awgu can comment.~~

## Before submitting
      Pull Request section?
      to it if that's the case.

cc @HamidShojanazeri


llama_finetuning.py
```diff
@@ -90,11 +95,26 @@ def main(**kwargs):
...
+        # model alone would consume 2+TB cpu mem (70 * 4 * 8)
+        if rank == 0:
+            model = LlamaForCausalLM.from_pretrained(
+                train_config.model_name,
+                load_in_8bit=True if train_config.quantization else None,
+                device_map="auto" if train_config.quantization else None,
+            )
+        else:
+            llama_config = LlamaConfig.from_pretrained(train_config.model_name)
+            with init_empty_weights():
```
awgu, line 110: Curious: Why do we need `init_empty_weights()` and not just `with torch.device("meta")`? What is insufficient with the latter?

llama_finetuning.py
```diff
@@ -127,14 +147,29 @@ def main(**kwargs):
 
         mixed_precision_policy, wrapping_policy = get_policies(fsdp_config, rank)
         my_auto_wrapping_policy = fsdp_auto_wrap_policy(model, LlamaDecoderLayer)
-   
+
+        # given the fast evolving PRs around meta device init, I am not sure
+        # what is the best param_init_fn atm, maybe we can switch to simple to_emtpy.
```
awgu, line 152: I think `to_empty(recurse=False)` seems okay.

lchu-ibm: @awgu thank you! I will make the change.

llama_finetuning.py
lchu-ibm: no specific reasons, originally doing it this way because the code is reading the model in HF way and specifically add `accelerate` in the requirements so thinking about using `accelerate.init_empty_weights` to match the model loading style.  Also, I remembered I hit some broadcast op issues when using with torch.device("meta") + sync_states in the past when this was introduced, let me double check it here and make the changes if looking good.

lchu-ibm: Interesting. If I switch to `with torch.device("meta")` while keeping my current `_param_init_fn` that I borrowed from one of your test cases, then it will fail with
```
NotImplementedError: Could not run 'c10d::broadcast_' with arguments from the 'Meta' backend
```

But, if I switch both, the combination of `with torch.device("meta")` + `to_empty(recurse=False)` works well. I will make both changes then.

lchu-ibm: addressed in https://github.com/facebookresearch/llama-recipes/pull/77/commits/1e64fc98d9296b28b7d34b11c28498f50af69e8b

llama_finetuning.py
lchu-ibm: addressed https://github.com/facebookresearch/llama-recipes/pull/77/commits/1e64fc98d9296b28b7d34b11c28498f50af69e8b

HamidShojanazeri (commented):
Thanks @lchu-ibm for the PR! pls refer to inline comment.

llama_finetuning.py
```diff
@@ -90,11 +93,26 @@ def main(**kwargs):
     gradient_accumulation_steps = train_config.batch_size_training // train_config.micro_batch_size
      
     # Load the pre-trained model and setup its configuration
-    model = LlamaForCausalLM.from_pretrained(
-        train_config.model_name,
-        load_in_8bit=True if train_config.quantization else None,
-        device_map="auto" if train_config.quantization else None,
-    )
+    if train_config.enable_fsdp:
```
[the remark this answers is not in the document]
lchu-ibm: @HamidShojanazeri makes sense.  So we don't have to force users use nightly if they do not use this feature.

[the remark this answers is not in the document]
lchu-ibm: I will address this.

llama_finetuning.py
```diff
@@ -127,14 +151,16 @@ def main(**kwargs):
...
-   
+
         model = FSDP(
             model,
             auto_wrap_policy= my_auto_wrapping_policy if train_config.use_peft else wrapping_policy,
             mixed_precision=mixed_precision_policy if not fsdp_config.pure_bf16 else None,
             sharding_strategy=fsdp_config.sharding_strategy,
             device_id=torch.cuda.current_device(),
             limit_all_gathers=True,
+            sync_module_states=True,
```
HamidShojanazeri, line 162: @lchu-ibm shall we add the check here as well, in case low_cpu and nightlies use this else None.

HamidShojanazeri:
Thanks @lchu-ibm for the updates, I would appreciate if we could add similar comments from the code to  [here](https://github.com/facebookresearch/llama-recipes/tree/main#fine-tuning-using-fsdp-only), and [here](https://github.com/facebookresearch/llama-recipes/blob/main/docs/mutli_gpu.md#fine-tuning-using-fsdp-only) as well.

llama_finetuning.py
lchu-ibm: https://github.com/facebookresearch/llama-recipes/pull/77/commits/c19c5c69aa4f2b0eb03119ef64c3aec164595ab9

llama_finetuning.py
```diff
@@ -62,8 +63,10 @@
 from torch.optim.lr_scheduler import StepLR
 from pkg_resources import packaging
 import torch
+import torch.nn as nn
 import torch.cuda.nccl as nccl
 import torch.distributed as dist
+from torch.distributed.fsdp._common_utils import _is_fsdp_flattened
```
chauhang, line 69: Is this being used anywhere in the code? Will be good to remove the unused imports, esp for the "_" methods which are internal and may change in future

llama_finetuning.py
```diff
@@ -83,25 +86,48 @@ def main(**kwargs):
         world_size = int(os.environ["WORLD_SIZE"])
 
     if torch.distributed.is_initialized():
-        torch.cuda.set_device(rank)
+        torch.cuda.set_device(local_rank)
```
chauhang, line 89: What is the reason for replacing "rank" with "local_rank"?

chauhang (changes requested):
@lchu-ibm Thanks for this PR to address the CPU OOM issues for 70B model. The code has some changes in the usage of "rank" and "local_rank". It will be good to test this on both single host multi-gpu and multi-host multi-gpu to verify things work correctly for both cases. It will be great if you can run the tests and attach the logs as well.

llama_finetuning.py
lchu-ibm: @chauhang I think the original usage of rank was a bug, maybe from copy-paste error or typo.  This function expects local rank so I fixed it.

This has been validated in two different ways:
1. I have validated running in both single node and multi node env, and can confirm the fix works but previous didn't work in multi-node env.  Same observation from others as discussed in https://github.com/facebookresearch/llama-recipes/issues/90
2. I also double checked other repos from pytorch and confirm the correctness and consistency of using local_rank, e.g. https://github.com/pytorch/workshops/blob/master/FSDP_Workshop/main_training.py#L382

lchu-ibm:
@chauhang Thanks for the suggestions! please see my inline response on the rank fix.  Also, I have just done a quick code cleanup by optimizing the imports as original code also have a bunch of unused imports.

HamidShojanazeri (approved):

llama_finetuning.py
lessw2020: we should be using local_rank here.  Rank is really the global rank.  
Local_rank = for intra-node (within node) assignments of gpu ids.

For single node, rank is the same numerically as local_rank and they are equivalent ids and will thus work. 

But for multi-node, rank is global_rank, unique across all nodes,  and will result in a failing gpu id within the second+ nodes if used in gpu device assignment since it would be trying to assign gpu ids higher than what the local node has.

llama_finetuning.py
```diff
@@ -83,25 +62,48 @@ def main(**kwargs):
...
+        # this avoids cpu oom when loading large models like llama 70B, in which case
+        # model alone would consume 2+TB cpu mem (70 * 4 * 8). This will add some comms
+        # overhead and currently requires latest nightly.
+        v = packaging.version.parse(torch.__version__)
+        verify_latest_nightly = v.is_devrelease and v.dev >= 20230701
+        if not verify_latest_nightly:
+            raise Exception("latest pytorch nightly build is required to run with low_cpu_fsdp config, "
+                            "please install latest nightly.")
+        if rank == 0:
+            model = LlamaForCausalLM.from_pretrained(
```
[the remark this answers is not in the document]
lchu-ibm: @rohan-varma for non-0 ranks, we are using `torch.device("meta")` init.

llama_finetuning.py
```diff
@@ -110,46 +112,49 @@ def main(**kwargs):
...
-   
+
         model = FSDP(
             model,
             auto_wrap_policy= my_auto_wrapping_policy if train_config.use_peft else wrapping_policy,
             mixed_precision=mixed_precision_policy if not fsdp_config.pure_bf16 else None,
             sharding_strategy=fsdp_config.sharding_strategy,
             device_id=torch.cuda.current_device(),
             limit_all_gathers=True,
+            sync_module_states=True if train_config.low_cpu_fsdp else False,
```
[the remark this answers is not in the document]
lchu-ibm: fixed. thanks for the catch!

pacman100:
Hello everyone, FYI: PR https://github.com/huggingface/transformers/pull/25107 and https://github.com/huggingface/accelerate/pull/1777 will handle the loading of models when using transformers to be efficient to avoid CPU OOMs when using FSDP **without any code changes from user side**. Currently, testing that out on 70B on single-node and multi-node setups.

HamidShojanazeri:
@pacman100 Thanks for the update, that would be very helpful will give it a try.

pacman100:
Hello, just using `AutoModelForCausalLM.from_pretrained()`  should work as long as one is using `accelerate` launcher with FSDP enabled. Basically, when FSDP is enabled with Accelerate, it sets env variable `ACCELERATE_USE_FSDP`  to `True`  and I am using that in the from_pretrained  method:

```
def is_fsdp_enabled():
    return strtobool(os.environ.get("ACCELERATE_USE_FSDP", "False")) == 1
```

So, if you don't want to use accelerate launcher, you can simply run `export ACCELERATE_USE_FSDP=true` and then have your own training loop wherein you properly use FDSP class with `sync_module_states=True`

How it works?
1. Have the model load on `meta` device on all ranks
2. Load the state dict only on rank==0 and set the param values from `meta` to `cpu` for rank==0
3. For all other ranks, do `torch.empty(*param.size(), dtype=dtype)` for every parameter on `meta` device
4. So, rank==0 will have loaded the model with correct state dict while all other ranks will have random/0 weights.
5. Set `sync_module_states=True` so that FSDP object takes care of broadcasting them to all the ranks before training starts.

here is the output on a 7B model:
```
accelerator.process_index=0 GPU Memory before entering the loading : 0
accelerator.process_index=0 GPU Memory consumed at the end of the loading (end-begin): 0
accelerator.process_index=0 GPU Peak Memory consumed during the loading (max-begin): 0
accelerator.process_index=0 GPU Total Peak Memory consumed during the loading (max): 0
accelerator.process_index=0 CPU Memory before entering the loading : 926
accelerator.process_index=0 CPU Memory consumed at the end of the loading (end-begin): 26415
accelerator.process_index=0 CPU Peak Memory consumed during the loading (max-begin): 31818
accelerator.process_index=0 CPU Total Peak Memory consumed during the loading (max): 32744
accelerator.process_index=0 model.lm_head.weight=Parameter containing:
tensor([[-0.0179,  0.0201, -0.0273,  ..., -0.0275, -0.0396, -0.0131],
        [-0.0510, -0.0079, -0.0383,  ..., -0.0481,  0.0581,  0.0282],
        [-0.0217, -0.0216, -0.0064,  ..., -0.0508,  0.0554, -0.0013],
        ...,
        [ 0.0425,  0.0452, -0.0131,  ...,  0.0019,  0.0476,  0.0342],
        [-0.0170, -0.0085,  0.0449,  ..., -0.0074,  0.0178,  0.0043],
        [-0.0439, -0.0859, -0.0820,  ...,  0.0130,  0.0669,  0.0884]],
       requires_grad=True)
accelerator.process_index=1 GPU Memory before entering the loading : 0
accelerator.process_index=1 GPU Memory consumed at the end of the loading (end-begin): 0
accelerator.process_index=1 GPU Peak Memory consumed during the loading (max-begin): 0
accelerator.process_index=1 GPU Total Peak Memory consumed during the loading (max): 0
accelerator.process_index=1 CPU Memory before entering the loading : 933
accelerator.process_index=1 CPU Memory consumed at the end of the loading (end-begin): 10
accelerator.process_index=1 CPU Peak Memory consumed during the loading (max-begin): 573
accelerator.process_index=1 CPU Total Peak Memory consumed during the loading (max): 1506
accelerator.process_index=1 model.lm_head.weight=Parameter containing:
tensor([[0., 0., 0.,  ..., 0., 0., 0.],
        [0., 0., 0.,  ..., 0., 0., 0.],
        [0., 0., 0.,  ..., 0., 0., 0.],
        ...,
        [0., 0., 0.,  ..., 0., 0., 0.],
        [0., 0., 0.,  ..., 0., 0., 0.],
        [0., 0., 0.,  ..., 0., 0., 0.]], requires_grad=True)
accelerator.process_index=0 GPU Memory before entering the prepare : 0
accelerator.process_index=0 GPU Memory consumed at the end of the prepare (end-begin): 13202
accelerator.process_index=0 GPU Peak Memory consumed during the prepare (max-begin): 15458
accelerator.process_index=0 GPU Total Peak Memory consumed during the prepare (max): 15458
accelerator.process_index=0 CPU Memory before entering the prepare : 27345
accelerator.process_index=0 CPU Memory consumed at the end of the prepare (end-begin): -26394
accelerator.process_index=0 CPU Peak Memory consumed during the prepare (max-begin): 0
accelerator.process_index=0 CPU Total Peak Memory consumed during the prepare (max): 27345
FullyShardedDataParallel(
  (_fsdp_wrapped_module): RWForCausalLM(
    (transformer): RWModel(
      (word_embeddings): Embedding(65024, 4544)
      (h): ModuleList(
        (0-31): 32 x FullyShardedDataParallel(
          (_fsdp_wrapped_module): DecoderLayer(
            (input_layernorm): LayerNorm((4544,), eps=1e-05, elementwise_affine=True)
            (self_attention): Attention(
              (maybe_rotary): RotaryEmbedding()
              (query_key_value): Linear(in_features=4544, out_features=4672, bias=False)
              (dense): Linear(in_features=4544, out_features=4544, bias=False)
              (attention_dropout): Dropout(p=0.0, inplace=False)
            )
            (mlp): MLP(
              (dense_h_to_4h): Linear(in_features=4544, out_features=18176, bias=False)
              (act): GELU(approximate='none')
              (dense_4h_to_h): Linear(in_features=18176, out_features=4544, bias=False)
            )
          )
        )
      )
      (ln_f): LayerNorm((4544,), eps=1e-05, elementwise_affine=True)
    )
    (lm_head): Linear(in_features=4544, out_features=65024, bias=False)
  )
)
accelerator.process_index=1 GPU Memory before entering the prepare : 0
accelerator.process_index=1 GPU Memory consumed at the end of the prepare (end-begin): 13202
accelerator.process_index=1 GPU Peak Memory consumed during the prepare (max-begin): 15458
accelerator.process_index=1 GPU Total Peak Memory consumed during the prepare (max): 15458
accelerator.process_index=1 CPU Memory before entering the prepare : 945
accelerator.process_index=1 CPU Memory consumed at the end of the prepare (end-begin): 4
accelerator.process_index=1 CPU Peak Memory consumed during the prepare (max-begin): 4
accelerator.process_index=1 CPU Total Peak Memory consumed during the prepare (max): 949
accelerator.process_index=1 model.lm_head.weight=Parameter containing:
tensor([[-0.0179,  0.0201, -0.0273,  ..., -0.0275, -0.0396, -0.0131],
        [-0.0510, -0.0079, -0.0383,  ..., -0.0481,  0.0581,  0.0282],
        [-0.0217, -0.0216, -0.0064,  ..., -0.0508,  0.0554, -0.0013],
        ...,
        [ 0.0425,  0.0452, -0.0131,  ...,  0.0019,  0.0476,  0.0342],
        [-0.0170, -0.0085,  0.0449,  ..., -0.0074,  0.0178,  0.0043],
        [-0.0439, -0.0859, -0.0820,  ...,  0.0130,  0.0669,  0.0884]],
       device='cuda:1', requires_grad=True)
accelerator.process_index=0 model.lm_head.weight=Parameter containing:
tensor([[-0.0179,  0.0201, -0.0273,  ..., -0.0275, -0.0396, -0.0131],
        [-0.0510, -0.0079, -0.0383,  ..., -0.0481,  0.0581,  0.0282],
        [-0.0217, -0.0216, -0.0064,  ..., -0.0508,  0.0554, -0.0013],
        ...,
        [ 0.0425,  0.0452, -0.0131,  ...,  0.0019,  0.0476,  0.0342],
        [-0.0170, -0.0085,  0.0449,  ..., -0.0074,  0.0178,  0.0043],
        [-0.0439, -0.0859, -0.0820,  ...,  0.0130,  0.0669,  0.0884]],
       device='cuda:0', requires_grad=True)
```

chauhang (approved):

PR #84, opened 2023-08-01
Coga

YosiMass:
make inference accept a file with several prompts

fix a bug in alpaca dataset that could truncate the response


YosiMass:
opened by mistake

PR #106, opened 2023-08-08
Fix finetuning batch size

rohan-mehta:
# Fix batch size in llama_finetuning.py

Currently, `llama_finetuning.py` uses `train_config.batch_size_training` as the batch size in the dataloader. This works fine right now, because the default config sets `batch_size_training` == `micro_batch_size` == `4`. But let's say you wanted an effecttive batch size of 128 and a micro batch size of 16. Then:
1. The dataloader would provide batches of 128 instead of 16
2. Gradient accumulation would happen ever (128 * 8) samples instead of every 128 samples
3. We would be loading 128 samples into memory each time, instead of 16

This PR fixes that issue.

Fixes # (issue)

I tested this by running the script with alpaca on an H100 - please let me know if theres anything more to be done.

## Before submitting
      Pull Request section?
      to it if that's the case.

Thanks for contributing 🎉!

### DIFF
--- a/llama_finetuning.py
+++ b/llama_finetuning.py
@@ -180,7 +180,7 @@ def main(**kwargs):
     # Create DataLoaders for the training and validation dataset
     train_dataloader = torch.utils.data.DataLoader(
         dataset_train,
-        batch_size=train_config.batch_size_training,
+        batch_size=train_config.micro_batch_size,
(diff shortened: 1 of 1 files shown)


gnohgnailoug:
I've also noticed this issue, but it seems like the problem still hasn't been resolved.

PR #122, opened 2023-08-16
Add FSDP CPU offloading option

helloworld1:
# What does this PR do?
Add FSDP CPU offloading option so the model can be trained on system with less GPU vram especially for training 70B models, which does not fit in 7xA100 GPUs without PEFT.

## Feature/Issue validation/testing
Use llama-2-70b-hf model without PEFT
```
torchrun --nnodes 1 --nproc_per_node 7  llama_finetuning.py --enable_fsdp --model_name /tmp/llama-2-70b-hf --dist_checkpoint_root_folder model_checkpoints --dist_checkpoint_folder fine-tuned 
```

Tested on without `--fsdp_cpu_offload`, it runs OOM on 7x A100 GPUs and crashes
With `--fsdp_cpu_offload` option, the training job can run fine without crashing

## Before submitting
- [ ] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).
- [x] Did you read the [contributor guideline](https://github.com/facebookresearch/llama-recipes/blob/main/CONTRIBUTING.md#pull-requests),
      Pull Request section?
- [ ] Was this discussed/approved via a Github issue? Please add a link
      to it if that's the case.
- [ ] Did you make sure to update the documentation with your changes?  
- [ ] Did you write any new necessary tests?

Thanks for contributing 🎉!

PR #162, opened 2023-09-03
Wandb, cpu offloading, cosine scheduler, CodeLlama, Completion and bug fixes

maximegmd:
# What does this PR do?

This PR contains a bit of everything I did (and will do) to improve my workflow, it can be merged but I am opening it with the intent of making these changes visible to users who are looking for them.

- Add FSDP CPU offloading for smaller configs #122 
- Add support for wandb tracking
- Add support for cosine + warmup scheduler
- Add support for CodeLlama
- Add support for completion corpus datasets with optional packing or padding

Fixes:
- Progress bar not incrementing correctly and displaying erroneous numbers.

- [x] LLama 7b full train on 4xV100 16GB
- [x] Llama 7b PEFT Lora on 4xV100 16GB
- [x] CodeLlama 7b full train on 4xV100 16GB

### DIFF
--- a/configs/training.py
+++ b/configs/training.py
@@ -30,6 +30,8 @@ class train_config:
     quantization: bool = False
     one_gpu: bool = False
     save_model: bool = True
+    use_cosine_scheduler: bool = False
+    warmup_steps: int = 10 # will be used if using cosing scheduler
     dist_checkpoint_root_folder: str="PATH/to/save/FSDP/model" # will be used if using FSDP
     dist_checkpoint_folder: str="fine-tuned" # will be used if using FSDP
     save_optimizer: bool=False # will be used if using FSDP
--- a/configs/wandb.py
+++ b/configs/wandb.py
@@ -1,0 +1,17 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# This software may be used and distributed according to the terms of the Llama 2 Community License Agreement.
+
+from dataclasses import dataclass, field
+
+@dataclass
+class wandb_config:
+    enable_wandb: bool = False
+    wandb_id: str = None
+    wandb_project: str = None
+    wandb_entity: str = None
+    wandb_mode: str = None
+    wandb_dir: str = None
+
+    
+    
+    
--- a/configs/wandb.py
+++ b/configs/wandb.py
@@ -11,7 +11,6 @@ class wandb_config:
     wandb_entity: str = None
     wandb_mode: str = None
     wandb_dir: str = None
-
-    
+    wandb_token: str = None
     
     
--- a/ft_datasets/__init__.py
+++ b/ft_datasets/__init__.py
@@ -4,3 +4,4 @@
 from .grammar_dataset import get_dataset as get_grammar_dataset
 from .alpaca_dataset import InstructionDataset as get_alpaca_dataset
 from .samsum_dataset import get_preprocessed_samsum as get_samsum_dataset
+from .completion_dataset import get_preprocessed_completion as get_completion_dataset
--- a/ft_datasets/completion_dataset.py
+++ b/ft_datasets/completion_dataset.py
@@ -1,0 +1,31 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# This software may be used and distributed according to the terms of the Llama 2 Community License Agreement.
+
+import datasets
+from transformers import DataCollatorWithPadding
+from .utils import Concatenator
+
+def get_preprocessed_completion(dataset_config, tokenizer, split):
+    if split == 'val':
+        return {}
+    
+    def prepare(sample):
+        sample = tokenizer(sample, padding=('max_length' if dataset_config.disable_packing else False))
+        source_ids = sample["input_ids"]
+        src_mask = sample["attention_mask"]
+
+        return {
+            "input_ids": source_ids,
+            "attention_mask": src_mask,
+            "labels": source_ids.copy(),
+        }
+    
+    dataset = datasets.load_dataset(dataset_config.data_path, split=split)
+    dataset = dataset.filter(lambda sample: sample["text"].strip() != '')
+    dataset = dataset.map(
+        lambda sample: prepare(sample["text"]),
+        batched=True
+    )
+    if not dataset_config.disable_packing:
+        dataset = dataset.map(Concatenator(chunk_size=4096), batched=True)
+    return dataset
--- a/llama_finetuning.py
+++ b/llama_finetuning.py
@@ -12,6 +12,7 @@ from pkg_resources import packaging
 from torch.distributed.fsdp import (
     FullyShardedDataParallel as FSDP,
 )
+from torch.distributed.fsdp.fully_sharded_data_parallel import CPUOffload
 from torch.optim.lr_scheduler import StepLR
 from torch.utils.data import DistributedSampler
 from transformers import (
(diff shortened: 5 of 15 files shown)


requirements.txt
```diff
@@ -9,8 +9,9 @@ black[jupyter]
 datasets
 fire
 git+https://github.com/huggingface/peft.git
-transformers>=4.31.0
+git+https://github.com/huggingface/transformers.git@main
```
mreso, line 12: Why are we going from release pypi to rolling release github here?

utils/train_utils.py
```diff
@@ -45,7 +46,7 @@ def set_tokenizer_params(tokenizer: LlamaTokenizer):
 def byte2mb(x):
     return int(x / 2**20)
 
-def train(model, train_dataloader,eval_dataloader, tokenizer, optimizer, lr_scheduler, gradient_accumulation_steps, train_config, fsdp_config=None, local_rank=None, rank=None):
+def train(model, train_dataloader,eval_dataloader, tokenizer, optimizer, lr_scheduler, gradient_accumulation_steps, train_config, fsdp_config=None, local_rank=None, rank=None, wandb_run=None, step_scheduler=False):
```
mreso, line 49: Do we need an extra parameter for step_scheduler? Can this not be lr_scheduler and we thread is differently based on type (if necessary)?

utils/train_utils.py
```diff
@@ -101,16 +104,25 @@ def train(model, train_dataloader,eval_dataloader, tokenizer, optimizer, lr_sche
...
+                            lr_scheduler.step()
                 else:
                     # regular backpropagation when fp16 is not used
                     loss.backward()
                     if (step + 1) % gradient_accumulation_steps == 0 or step == len(train_dataloader) - 1:
                         optimizer.step()
                         optimizer.zero_grad()
-                        pbar.update(step//gradient_accumulation_steps)
+                        pbar.update()
+                        if step_scheduler:
```
mreso, line 118: Is it relevant to update scheduler on different places? Would one not be sufficient and you just adapt the step length? Like we call update every step instead of every second but double step size?

utils/train_utils.py
```diff
@@ -396,8 +412,6 @@ def save_train_params(train_config, fsdp_config, rank):
     train_config.dist_checkpoint_root_folder
     + "/"
     + train_config.dist_checkpoint_folder
-    + "-"
```
mreso, line 399: Why this change?

configs/training.py
```diff
@@ -30,6 +30,8 @@ class train_config:
     quantization: bool = False
     one_gpu: bool = False
     save_model: bool = True
+    use_cosine_scheduler: bool = False
```
mreso, line 33: There are many different schedulers, can we make this an easily extensible enum or similar to avoid having n use_X_scheduler bools? Given that all need different parameters a separate config dataclass for schedulers would work as well and fit our pattern. Compare datasets

configs/wandb.py
```diff
@@ -0,0 +1,16 @@
...
+from dataclasses import dataclass, field
+
+@dataclass
+class wandb_config:
+    enable_wandb: bool = False
+    wandb_id: str = None
+    wandb_project: str = None
+    wandb_entity: str = None
+    wandb_mode: str = None
+    wandb_dir: str = None
```
mreso, line 13: Prefix is not necessary as its inside wandb_config. You will configure it through --wandb_config.dir from the command line instead of --wandb_config.wandb_dir. Could we populate some of these (like dir) with meaningful default values?

configs/wandb.py
```diff
@@ -0,0 +1,16 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# This software may be used and distributed according to the terms of the Llama 2 Community License Agreement.
+
+from dataclasses import dataclass, field
+
+@dataclass
+class wandb_config:
+    enable_wandb: bool = False
```
mreso, line 8: We should move enable wandb in training_config and only then read in this config.

ft_datasets/__init__.py
```diff
@@ -3,4 +3,5 @@
 
 from .grammar_dataset import get_dataset as get_grammar_dataset
 from .alpaca_dataset import InstructionDataset as get_alpaca_dataset
-from .samsum_dataset import get_preprocessed_samsum as get_samsum_dataset
\ No newline at end of file
+from .samsum_dataset import get_preprocessed_samsum as get_samsum_dataset
+from .completion_dataset import get_preprocessed_completion as get_completion_dataset
```
mreso, line 7: Can we rename completion to code completion to make it more obvious to the user?

llama_finetuning.py
```diff
@@ -12,21 +12,23 @@
 from torch.distributed.fsdp import (
     FullyShardedDataParallel as FSDP,
 )
+from torch.distributed.fsdp.fully_sharded_data_parallel import CPUOffload
```
mreso, line 15: You can import CPUOffload directly from torch.distributed.fsdp together in line 12

llama_finetuning.py
```diff
@@ -12,21 +12,23 @@
...
 from torch.optim.lr_scheduler import StepLR
 from torch.utils.data import DistributedSampler
 from transformers import (
     LlamaForCausalLM,
-    LlamaTokenizer,
+    AutoTokenizer,
```
mreso, line 20: Why do we change this to AutoTokenizer?

utils/wandb_utils.py
```diff
@@ -0,0 +1,16 @@
+import os
+from configs import wandb_config
+
+def wandb_watch(model, cfg: wandb_config):
+    import wandb
```
mreso, line 5: As we're importing wandb only if we enable it we could potentially make it an optional dependency (compare vllm in pyproject.toml). Though I am not sure if its not installed as a dependency already.

ft_datasets/completion_dataset.py
```diff
@@ -0,0 +1,31 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
```
mreso, line 1: Would be good to integrate a test for the dataset.

utils/train_utils.py
```diff
@@ -101,16 +104,25 @@ def train(model, train_dataloader,eval_dataloader, tokenizer, optimizer, lr_sche
                         scaler.step(optimizer)
                         scaler.update()
                         optimizer.zero_grad()
-                        pbar.update(step//gradient_accumulation_steps)
+                        pbar.update()
+                        if step_scheduler:
```
mreso, line 108: see below

mreso (changes requested):
Thanks for the contribution! Left a few comments. Please also address the conflicts.

requirements.txt
maximegmd: The fast code llama tokenizer wasn't working correctly and only the git version had the fix at the time of this change, hopefully it can be rolled back now

utils/train_utils.py
maximegmd: When passing full paths as the model's name it creates chaotic paths

ft_datasets/__init__.py
maximegmd: it's not code completion, it's just corpus text completion, it's what I use for pretraining from textbooks for example

llama_finetuning.py
maximegmd: To support the CodeLLamaTokenizer automatically

PR #165, opened 2023-09-05
Pass `use_cache=False` when training with FSDP

awgu:
# What does this PR do?

This PR passes `use_cache=False` if using FSDP (since the default is [`True`](https://github.com/huggingface/transformers/blob/aea761499f4b1193f2706f471442da6f9df65d65/src/transformers/models/llama/configuration_llama.py#L120)). The current implementation assumes that FSDP is only used for training (i.e. not inference), in which case we do not need to cache past key/value states. For the LLaMA-7B run, this saves ~4 GB in activation memory and avoids peak memory from being at the end of the forward pass without affecting the end-to-end time.

<details>
<summary> Before (use_cache=True) </summary>

snapshot_7b_perparam_8gpus
Note the peak memory at the end of forward. This happens because we are saving past key/value states that are not needed for gradient computation.

- Peak active: 14.93 GB
- Peak reserved: 19.52 GB
(The numbers from a prototype FSDP implementation, but existing FSDP shows a similar decrease since we only change the activation memory.)

</details>

<details>
<summary> After (use_cache=False) </summary>

snapshot_7b_perparam_8gpus

- Peak active: 10.94 GB
- Peak reserved: 15.11 GB

</details>

```
torchrun --nnodes 1 --nproc_per_node 8  llama_finetuning.py --enable_fsdp --model_name llama-2-7b-hf/ --pure_bf16 --output_dir output_dir --use_fast_kernels
```

Feel free to take over this PR and test further.

### DIFF
--- a/llama_finetuning.py
+++ b/llama_finetuning.py
@@ -66,6 +66,7 @@ def main(**kwargs):
         setup_environ_flags(rank)
 
     # Load the pre-trained model and setup its configuration
+    use_cache = False if train_config.enable_fsdp else None
     if train_config.enable_fsdp and train_config.low_cpu_fsdp:
         """
         for FSDP, we can save cpu memory by loading pretrained model on rank0 only.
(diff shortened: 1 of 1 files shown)


llama_finetuning.py
```diff
@@ -94,6 +97,7 @@ def main(**kwargs):
             train_config.model_name,
             load_in_8bit=True if train_config.quantization else None,
             device_map="auto" if train_config.quantization else None,
+            use_cache=use_cache,
```
awgu, line 100: So far, I have only tested this code path (not the `low_cpu_fsdp=True` path). If you guys have bandwidth, feel free to cherry pick this and test.

chauhang (approved):

PR #203, opened 2023-09-15
adding optimizer overlap for FSDP

HamidShojanazeri:
# What does this PR do?

This PR adds Optimizer overlap that bring addition memory savings by fusing the gradinet calculation and parameter update steps.

for 7B --> max reserved memory saving is 7% 
          ---> allocated memory 4%
         ---> active memory 4%

- [ ] Test A without optimizer_overlap
https://gist.github.com/HamidShojanazeri/08ee3d23bdb0fa60466071dee1efda1f

- [ ] Test B with optimizer_overlap
https://gist.github.com/HamidShojanazeri/3d1147012e9db130dd7cebf75d3caa64

[Logs with/out anyprecision ](https://gist.github.com/HamidShojanazeri/a8b3689d012f657ef8e84baa4613c13c)

[Logs with AdamW](https://gist.github.com/HamidShojanazeri/cd1a705ecbf238633248a122e258531d)

## Before submitting
      Pull Request section?
- [ X] Was this discussed/approved via a Github issue? Please add a link
      to it if that's the case.
- [X] Did you make sure to update the documentation with your changes?  

Thanks for contributing 🎉!


README.md
```diff
@@ -144,6 +144,14 @@ Setting `use_fast_kernels` will enable using of Flash Attention or Xformer memor
 torchrun --nnodes 1 --nproc_per_node 4  examples/finetuning.py --enable_fsdp --use_peft --peft_method lora --model_name /patht_of_model_folder/7B --pure_bf16 --output_dir Path/to/save/PEFT/model --use_fast_kernels
 ```
 
+## FSDP optimizer overlap
+
+setting `optimizer_overlap` in [fsdp_config](./src/llama_recipes/configs/fsdp.py) enable optimizer overlap that lowers the GPU memory footprint during the training. The main idea here is the fusion of gradient calculation and parameter update in the one step. This is a new feature in FSDP and is only available on PyTorch nightly binaries for versions before 2.1.0.
```
chauhang, line 149: What is the reason for stating "for versions before 2.1.0"?

src/llama_recipes/finetuning.py
```diff
@@ -228,12 +238,29 @@ def main(**kwargs):
             use_kahan_summation=False,
             weight_decay=train_config.weight_decay,
         )
+    elif optimizer_in_backward_available:
+        print(f"setting up optimizer overlap")
+        optim_kwargs = {"lr": train_config.lr}
+        _apply_optimizer_in_backward(
+            optimizer_class=optim.AdamW,
```
chauhang, line 245: If the optimizer_in_backward_available flag is set and user has selected AnyPrecisionAdamW, will be good to add for that case as well, unless there is a restriction on which optimizers support this feature.

src/llama_recipes/finetuning.py
```diff
@@ -228,12 +238,29 @@ def main(**kwargs):
             use_kahan_summation=False,
             weight_decay=train_config.weight_decay,
         )
+    elif optimizer_in_backward_available:
+        print(f"setting up optimizer overlap")
+        optim_kwargs = {"lr": train_config.lr}
```
chauhang, line 243: Please see if setting foreach=False param for the optimizer will bring further memory savings

chauhang (changes requested):
@HamidShojanazeri Thanks for this PR. Left few comments. It will be great to capture a memory profile as well

README.md
rohan-varma, line 149: versions including and after 2.1.0?

src/llama_recipes/finetuning.py
```diff
@@ -228,12 +238,29 @@ def main(**kwargs):
...
+        _apply_optimizer_in_backward(
+            optimizer_class=optim.AdamW,
+            params=model.parameters(),
+            optimizer_kwargs=optim_kwargs,
+            register_hook=False,
+        )
+        for p in model.parameters():
+            assert hasattr(p, "_in_backward_optimizers")
+        optim_kwargs = {"lr": train_config.lr, "weight_decay":0.0}
```
rohan-varma, line 252: This logic is sort of weird, because we redefine optim_kwrags, then define optimizer, also its defined in the else, can we consolidate?

src/llama_recipes/finetuning.py
rohan-varma: @chauhang foreach may help issue less CUDA kernel calls overall and reduce kernel launch overhead, but I don't see how it would result in more memory savings.

src/llama_recipes/finetuning.py
HamidShojanazeri: Added for Anyprecision as well, as per my test it does not break.

rohan-varma (approved):

PR #207, opened 2023-09-17
adding chat special tokens to tokenizer

jveronvialard:
# What does this PR do?

Small refactoring to add chat-based special tokens B_INST, E_INST, B_SYS, E_SYS to tokenizer, and then resizing model embedding matrix accordingly (not using pad_to_multiple_of to avoid dependency on https://github.com/huggingface/transformers/pull/25732). This follows implementation from StarCoder: https://github.com/bigcode-project/starcoder/blob/main/chat/train.py#L132.

Updated `test_custom_dataset()` from `test_custom_dataset.py` accordingly. Also tested by running `src.llama_recipes.finetuning.py`.

## Before submitting
      Pull Request section?
      to it if that's the case.
- [x] Did you write any new necessary tests?

Thanks for contributing 🎉!


jveronvialard:
@mreso any thoughts? happy to do some refactoring if needed

mreso:
> @mreso any thoughts? happy to do some refactoring if needed

Thanks for the contribution! Looks good to me overall. You did move some code upwards though (dataset creation etc) which is fine as the functionality is not affected but we've a bigger pr in the pipeline which will introduce a bunch of changes to finetuning.py as well. Merging this will be a nightmare if we merge this on as is. Could you refactor so only the essential parts are included? Otherwise we can also wait until we have #206 in.

jveronvialard:
@mreso is it better now? I think I need to add the special tokens (done when `get_preprocessed_dataset()` is called) and resize the model embeddings before the model gets wrapped to FSDP

PR #250, opened 2023-10-12
Introduce a tracker API and integrate Aimstack tracker

dushyantbehl:
# What does this PR do?

This pull request adds a tracker API to llama recipes and integrates Aimstack https://aimstack.io/
The goal of tracker API is to create a standard template to attach any other experiment trackers.

With aim integration there is a dependency on the `aim` package which needs to be installed separately before using this.

Fixes # (issue)

Relates to issue https://github.com/facebookresearch/llama-recipes/issues/225

#Before submitting
      Pull Request section?
- [x] Was this discussed/approved via a Github issue? Please add a link
      to it if that's the case.
      https://github.com/facebookresearch/llama-recipes/issues/225

Thanks for contributing 🎉!


dushyantbehl:
Opening up the PR for initial review. Feature has to be fully tested and unit test and documentation fixed.

dushyantbehl:
@HamidShojanazeri @nairbv opening this PR for feedback. Sorry it has been long and I have been occupied with other things. 
I have added a tracker API which can be used to add support for any tracker like WandB / Aim along with Aim integration.

docs/experiment_tracking.md
```diff
@@ -0,0 +1,30 @@
+# Experiment tracking in llama recipes. 
+
+Llama recipes has an optional support for tracking experiments using experiment tracker modules like Aimstack ().
+
+The experiment tracker can be enabled by passing an additional argument to the llama recipes code which loads and
+tracks all the training and evaluation metrics in the tracker along with the system metrics being tracker by these trackers.
+
+The addition argument can be passed as -
```
nairbv, line 8: missing the argument? looks like the below might be a placeholder?

requirements.txt
```diff
@@ -12,4 +12,5 @@ transformers>=4.31.0
 sentencepiece
 py7zr
 scipy
-optimum
\ No newline at end of file
+optimum
+aim=3.7.11
```
nairbv, line 16: probably want to make this optional. I think we're going to want to ensure that users that don't have (don't want) aim installed should still be able to use llama-recipes.

src/llama_recipes/utils/train_utils.py
```diff
@@ -66,6 +67,11 @@ def train(model, train_dataloader,eval_dataloader, tokenizer, optimizer, lr_sche
     checkpoint_times = []
     results = {}
     best_val_loss = float("inf")
+
+    track_metrics = True
+    if train_config.enable_fsdp and rank != 0:
+        track_metrics = False
```
nairbv, line 73: just `track_metrics = ....` in one line is a bit cleaner. something like `track_metrics = not fsdp or rank == 0` I think would be equivalent.

src/llama_recipes/utils/train_utils.py
```diff
@@ -114,24 +123,27 @@ def train(model, train_dataloader,eval_dataloader, tokenizer, optimizer, lr_sche
...
+
+        if tracker is not None and track_metrics:
+            tracker.track(epoch_end_time , name='epoch_times', stage='train')
+            tracker.track(train_epoch_loss , name='train_loss', stage='train')
+            tracker.track(train_perplexity , name='train_perplexity', stage='train')
+            tracker.track(memtrace.peak, name="Max CUDA Mem Allocated(GB)", stage='train')
+            tracker.track(memtrace.max_reserved, "Max CUDA Mem Reserved(GB)", stage='train')
+            tracker.track(memtrace.peak_active_gb, "Peak active CUDA Mem(GB)", stage='train')
+            tracker.track(memtrace.cuda_malloc_retires, "Cuda Malloc retires", stage='train')
+            tracker.track(memtrace.cpu_peaked + memtrace.cpu_begin, "CPU Total Peak Mem Max(GB)", stage='train')
```
nairbv, line 142: It looks a bit awkward having two very similar pieces of code logging the same info two different ways.

Would it make sense to combine print logging above with the 'tracker' here somehow? e.g. maybe an alternate default implementation of tracker=PrintTracker that just prints. Or maybe the set of things to be logged can be passed to both the tracker and print statements in a loop.

src/llama_recipes/utils/train_utils.py
```diff
@@ -179,17 +191,14 @@ def train(model, train_dataloader,eval_dataloader, tokenizer, optimizer, lr_sche
...
+            if tracker is not None and track_metrics:
+                tracker.track(best_val_loss , name='best_val_loss', stage='validation')
+                tracker.track(eval_epoch_loss , name='eval_epoch_loss', stage='validation')
             val_loss.append(best_val_loss)
             val_prep.append(eval_ppl)
-        if train_config.enable_fsdp:
-            if rank==0:
-                print(f"Epoch {epoch+1}: train_perplexity={train_perplexity:.4f}, train_epoch_loss={train_epoch_loss:.4f}, epoch time {epoch_end_time}s")
-        else:
+        if track_metrics:
```
nairbv, line 201: it looks like this change is removing the logging of epoch loss and time in the fsdp case?

src/llama_recipes/utils/train_utils.py
```diff
@@ -98,10 +104,13 @@ def train(model, train_dataloader,eval_dataloader, tokenizer, optimizer, lr_sche
                         optimizer.step()
                         optimizer.zero_grad()
                         pbar.update(1)
+                if tracker is not None and track_metrics:
+                    tracker.track(loss, name='loss', stage='train')
```
nairbv, line 108: should we be gathering loss from all ranks in the fsdp case?

nairbv: also, can we track throughput (tokens per second and/or tokens per step) during training?

nairbv:
looks like there are conflicts with main that will have to be resolved too

src/llama_recipes/utils/train_utils.py
dushyantbehl: The original code checks if fsdp is enabled and if so prints only on rank 0. if fsdp is not enabled it prints anyway.

`track_metrics` is used as a short cut to check this condition. It is set as `not (train_config.enable_fsdp and rank != 0)` to mimic this condition and remove a duplicate print statement.

src/llama_recipes/utils/train_utils.py
dushyantbehl: Our team's patch which fixes GPU metrics tracking is merged in Aim now. We do not need extra memory tracking I think.

dushyantbehl: Keeping the interface simple and updated the code. Any extra can be done as additonal patch in my opinion.

src/llama_recipes/utils/train_utils.py
dushyantbehl: The original code tracked only on rank 0 in fsdp. 
If we need to track on all ranks in Aim, we can do that.

src/llama_recipes/utils/train_utils.py
dushyantbehl: done

src/llama_recipes/utils/train_utils.py
nairbv: ah maybe I misread the condition the first time

src/llama_recipes/utils/train_utils.py
```diff
@@ -207,6 +209,17 @@ def train(model, train_dataloader,eval_dataloader, tokenizer, optimizer, lr_sche
     results["avg_epoch_time"] = avg_epoch_time
     results["avg_checkpoint_time"] = avg_checkpoint_time
 
+    if tracker is not None and track_metrics:
+        tracker.track(avg_train_loss , name='avg_train_loss', context={'subset':'experiment'})
```
nairbv, line 213: wrong parameter names? let's make sure we test this to the end of an epoch. we can truncate a dataset so it runs to completion faster.

dushyantbehl: Fixed. Tested and updated the documentation.

docs/experiment_tracking.md
```diff
@@ -0,0 +1,88 @@
+# Experiment tracking in llama recipes. 
+
+Llama recipes has an optional support for tracking experiments using experiment tracker modules like Aimstack or WandB (Currently not supported).
```
nairbv, line 3: nit:
I'd drop 'an' in 'an optional'

I think "Currently only Aimstack is supported" might be clearer, "not supported' doesn't make it clear what is/isn't.

docs/experiment_tracking.md
```diff
@@ -0,0 +1,88 @@
...
+Current version of the code supports Aimstack tracking via the trackers module.
+
+One needs to install Aimstack separately if need to use aimstack as a tracker.
+
+### Installation
+
+Aim can be installed by running
+
+```
+$ pip install aim==3.17.5
```
nairbv, line 30: do we require a specific version or can we put `>=` ?

docs/experiment_tracking.md
```diff
@@ -0,0 +1,88 @@
...
+```
+
+### Aim setup
+
+Aimstack is a self hosted open source component so a small setup is needed before running Aimstack.
```
nairbv, line 35: nit: s/a small/some/

docs/experiment_tracking.md
```diff
@@ -0,0 +1,88 @@
...
+           --model_name $MODEL_NAME --output_dir $OUTPUT_DIR \
+           --use_peft --peft_method lora --pure_bf16 \
+           --enable_fsdp --use_fast_kernels --num_epochs 1 --batch_size_training 1\
+           --tracker aim --aim_config.repo $AIMSTACK_DB \
+           --aim_config.experiment llama7b-finetuning
+```
+
+### Example screenshots.
+
+This is how the runs would look like,
```
nairbv, line 80: nit: either "how the runs look" or "what the runs look like" not "how the runs look like"

When referring to a code snippet or image, it might also be better to be consistent in terms of punctuation. some end in a comma, some a period, some nothing. Maybe a colon would be better?

src/llama_recipes/utils/train_utils.py
```diff
@@ -207,6 +209,17 @@ def train(model, train_dataloader,eval_dataloader, tokenizer, optimizer, lr_sche
...
+    if tracker is not None and track_metrics:
+        tracker.track(avg_train_loss , name='avg_train_loss', stage={'subset':'experiment'})
+        tracker.track(avg_train_prep, name='avg_train_prep', stage={'subset':'experiment'})
+
+        if train_config.run_validation:
+            tracker.track(avg_eval_prep , name='avg_eval_prep', stage={'subset':'validation'})
+            tracker.track(avg_eval_loss, name='avg_eval_loss', stage={'subset':'validation'})
+
+        tracker.track(avg_epoch_time , name='avg_epoch_time', stage={'subset':'experiment'})
+        tracker.track(avg_checkpoint_time , name='avg_checkpoint_time', stage={'subset':'experiment'})
```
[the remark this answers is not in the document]
dushyantbehl: yes this was fixed. thanks.

docs/experiment_tracking.md
dushyantbehl: Aimstack sdk is changing across versions so I am putting the version which I have found to be stable.

src/llama_recipes/configs/trackers.py
```diff
@@ -0,0 +1,16 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# This software may be used and distributed according to the terms of the Llama 2 Community License Agreement.
+
+from dataclasses import dataclass
+
+@dataclass
+class aim_config:
+    experiment: str = "llama-recipes"
+    # This is either the location of a locallly accessible directory example `.run`
```
nairbv, line 9: nit: seems like there's some punctuation missing?

src/llama_recipes/configs/trackers.py
```diff
@@ -0,0 +1,16 @@
...
+    # or can be the location of a remote repository which is hosted on a server.
+    # if remote_server_ip or remote_server_port is set then the repo will be set to
+    # remote aim repo, else the directory specified by `repo` takes precedence.
+    # The directory defaults to None which means `.run` in aim.
```
nairbv, line 13: is the default `~/.run` or really just `.run` ?

nairbv (approved):
Some minor comments but overall LGTM.

We probably need to have someone from Meta who has committed to this repo take a look before merging. @HamidShojanazeri / @WuhanMonkey  / @jeffxtang  ?

dushyantbehl:
@nairbv @HamidShojanazeri @chauhang I noticed a spell check was failing for this PR which is fixed now. Please take a look and let me know if anything else needs to be fixed.

PR #279, opened 2023-11-01
Llama 2 On-Prem Inference Using vLLM and TGI

jeffxtang:
# What does this PR do?

This tutorial shows how to use Llama 2 with vLLM and Hugging Face TGI and how to use LangChain to talk to vLLM and TGI hosted Llama 2 on prem.

Fixes # (issue)

- [ ] Test A
Logs for Test A

- [ ] Test B
Logs for Test B

## Before submitting
- [X ] Did you read the [contributor guideline](https://github.com/facebookresearch/llama-recipes/blob/main/CONTRIBUTING.md#pull-requests),
      Pull Request section?
      to it if that's the case.
- [X ] Did you make sure to update the documentation with your changes?  

Thanks for contributing 🎉!


demo_apps/llama-on-prem.md
```diff
@@ -0,0 +1,180 @@
...
+
+## Setting up vLLM with Llama 2
+
+On a terminal, run the following commands:
+```
+conda create -n vllm python=3.8
+conda activate vllm
+pip install vllm
+cd <your_work_folder>
+git clone https://github.com/vllm-project/vllm
```
WuhanMonkey, line 21: If we already pip install the vllm package, do we still need to clone the source repo for building locally? Maybe we can add some context here stating there are two ways to install vllm.

demo_apps/llama-on-prem.md
```diff
@@ -0,0 +1,180 @@
...
+cd vllm/vllm/entrypoints/
+```
+
+Then run `huggingface-cli login` and copy and paste your Hugging Face access token to complete the login.
+
+There are two ways to deploy Llama 2 via vLLM, as a general API server or an OpenAI-compatible server.
```
WuhanMonkey, line 27: Just a quick note that Azure API hosting schema will also be using the one similar to OpenAI. Might be helpful to link documentation page here - https://platform.openai.com/docs/api-reference/chat/object

demo_apps/llama-on-prem.md
```diff
@@ -0,0 +1,180 @@
+# Llama 2 On-Prem Inference Using vLLM and TGI
+
+Enterprise customers may prefer to deploy Llama 2 on-prem and run Llama in their own servers. This tutorial shows how to use Llama 2 with [vLLM](https://github.com/vllm-project/vllm) and Hugging Face [TGI](https://github.com/huggingface/text-generation-inference), two leading open-source tools to deploy and serve LLMs, and how to create vLLM and TGI hosted Llama 2 instances with [LangChain](https://www.langchain.com/), an open-source LLM app development framework which we used for our earlier demo apps with Llama 2 running on [local Mac](https://github.com/facebookresearch/llama-recipes/blob/main/demo_apps/HelloLlamaLocal.ipynb) or [Replicate cloud](https://github.com/facebookresearch/llama-recipes/blob/main/demo_apps/HelloLlamaCloud.ipynb). See [here](https://medium.com/@rohit.k/tgi-vs-vllm-making-informed-choices-for-llm-deployment-37c56d7ff705) for a detailed comparison of vLLM and TGI.
+
+We'll use the Amazon EC2 instance running Ubuntu with an A10G 24GB GPU as an example of running vLLM and TGI with Llama 2, and you can replace this with your own server to implement on-prem Llama 2 deployment.
```
WuhanMonkey, line 5: Have we ever consider using 2xA10G GPU here? In that case, we can also provide more advanced tutorial on model sharding or running two models on one instance by utilizing specific CUDA cores etc.

demo_apps/llama-on-prem.md
```diff
@@ -0,0 +1,180 @@
...
+Run the command below to deploy vLLM as a general Llama 2 service:
+
+```
+python api_server.py --host 0.0.0.0 --port 5000 --model meta-llama/Llama-2-7b-chat-hf
+```
+
+Then on another terminal you can run:
+
+```
+curl http://localhost:5000/generate -d '{
```
WuhanMonkey, line 40: Besides the CLI, should we provide a python example snippet as well?

demo_apps/llama-on-prem.md
jeffxtang: actually clone the vllm repo is not for building locally, but for using its scripts like api_server.py

and clone without "pip install vllm" won't work either, because dependency packages such as fastapi need to be installed to run the script.

but yes, i'll add a note to avoid the possible misunderstanding.

demo_apps/llama-on-prem.md
jeffxtang: ok i added a note, but with the OpenAI API authentication link - the link above is not related to how the OpenAI compatible LLM instance is created.

demo_apps/llama-on-prem.md
jeffxtang: I showed a command later to run vLLM on 4 GPUs with a new argument. This tutorial is still one of the quick start guides. I don't know how much time I'll need to add the sharding etc content. If you or @HamidShojanazeri has already done something like this, please add to the PR or maybe in a new PR.

demo_apps/llama-on-prem.md
jeffxtang: The Google Colab shows complete Python code to talk to the servers. The curl commands are just for quick verification.

demo_apps/llama-on-prem.md
WuhanMonkey: Can you double check to see if we really need the vllm repo here? I do see the way you launch api server a bit different from mine. Here is how I did it: `python -m vllm.entrypoints.openai.api_server  --model meta-llama/Llama-2-70b-chat-hf --tensor-parallel-size 8 --disable-log-requests --port  8000`

demo_apps/llama-on-prem.md
WuhanMonkey: OK, I did see in the colab we inferenced the API server via Langchain. I wonder if we want to add some example using POST in python, which is more raw. I have some example, let me link it later.

WuhanMonkey: The reason is that vllm documentation is not so clear about this part. If we can add some examples here, I think it would helpful

WuhanMonkey: Here is an [example](https://github.com/WuhanMonkey/Azure_API_hosting_benchmark/blob/main/vm-local/70b_chat_vllm_benchmark.py#L148) how I used the urlib to post requests to the api server.

demo_apps/llama-on-prem.md
jeffxtang: Yes I tested in a new conda env without "pip install vllm" and then running "python api_server.py..." caused the error of unable to find fastapi. Your way works because you don't cd to vllm/entrypoints/openai. My way works because I cd to vllm/vllm/entrypoints first. All imports of vllm's source don't require "pip install vllm" but other packages required by vllm need to be installed first. 

You have to use one of the two ways to install first before running the api_server.py: https://vllm.readthedocs.io/en/latest/getting_started/installation.html

demo_apps/llama-on-prem.md
jeffxtang: Thanks but your examples seems to be in your private repo so I can't open it. 

Converting the curl command to a Python script is really straightforward, that's why I had the content "Now in your Llama client app, you can make an HTTP request as the `curl` command above to send a query to Llama and parse the response." I assume if one needs that script, he/she can simply ask ChatGPT like this :)
https://chat.openai.com/c/013cc516-71db-4ee7-b8e6-3ec946734cfe

demo_apps/llama-on-prem.md
WuhanMonkey: Correct, I think that is what I mean here. You either run `pip install vllm` and install all dependencies as part of pip packages or install from source then run `pip install requirements.txt` or `pip install -e .` in the repo to install dependencies. That would avoid mis-match in the dependency package versions. There are chances where the dependencies versions in the main repo is different from pip packages.

demo_apps/llama-on-prem.md
WuhanMonkey: Sure, I can add a note in this PR for launching model replicas. Will add around your command which set TP for 4 GPUs

demo_apps/llama-on-prem.md
jeffxtang: Also, I only showed the creation of the Llama instance "llm" in the Colab - but using LangChain or LlamaIndex will quickly be a lot more powerful and effective, when building Llama RAG apps, than using the raw post request which is mainly for quick API test.

WuhanMonkey: Can't access your ChatGPT result :) it doesn't load for me. I have invited you to that repo, so you can take a look. Sure. I think it is a different use case, especially when you are running benchmarks and doesn't want to use any framework. Might not a good fit in this PR.

demo_apps/llama-on-prem.md
```diff
@@ -0,0 +1,181 @@
...
+        "temperature": 0
+    }'
+```
+
+Also, if you have multiple GPUs, you can add the `--tensor-parallel-size` argument when starting the server (see [here](https://vllm.readthedocs.io/en/latest/serving/distributed_serving.html) for more info). For example, the command below runs the Llama 2 13b-chat model on 4 GPUs:
+
+```
+python api_server.py --host 0.0.0.0 --port 5000 --model meta-llama/Llama-2-13b-chat-hf --tensor-parallel-size 4
+```
+
```
WuhanMonkey, line 69:
```suggestion
With multiple GPUs, you can also run replica of models as long as your model size can fit into targeted GPU memory. For example, if you have two A10G with 24 GB memory, you can run two 7B Llama 2 models at the same time. This can be done by launching two api servers each targeting specific CUDA cores on different ports:
`CUDA_VISIBLE_DEVICES=0 python api_server.py --host 0.0.0.0 --port 5000  --model meta-llama/Llama-2-7b-chat-hf --tensor-parallel-size 1 `
and
`CUDA_VISIBLE_DEVICES=1 python api_server.py --host 0.0.0.0 --port 5001  --model meta-llama/Llama-2-7b-chat-hf --tensor-parallel-size 1 `
The benefit would be now you can balance incoming requests to both models, reaching higher batch size processing for a trade-off of generation latency.


```

demo_apps/llama-on-prem.md
WuhanMonkey: Added the suggestion below

demo_apps/llama-on-prem.md
jeffxtang: So back to your original question "If we already pip install the vllm package, do we still need to clone the source repo for building locally?" -  the answer is yes so the commands there are correct. I also added a note to clarify.

demo_apps/llama-on-prem.md
jeffxtang: Can you access the link now? https://chat.openai.com/share/d289e3f8-9d84-4eea-b4d0-fafd752150c9 (I should've clicked the Share button first). If you still can't, you can just enter in ChatGPT this:

write a python script that implements the curl command below:

curl http://localhost:5000/generate -d '{
        "prompt": "Who wrote the book Innovators dilemma?",
        "max_tokens": 300,
        "temperature": 0
    }'

demo_apps/llama-on-prem.md
jeffxtang: Thanks for the note - how do the two commands differ from this single one?

python api_server.py --host 0.0.0.0 --port 5000  --model meta-llama/Llama-2-7b-chat-hf --tensor-parallel-size 2

and this one?

CUDA_VISIBLE_DEVICES=0,1 python api_server.py --host 0.0.0.0 --port 5000  --model meta-llama/Llama-2-7b-chat-hf --tensor-parallel-size 2

WuhanMonkey: Both of your command is running tensor parallelsim. Basically sharding the compute into two GPUs and merge the results together. Whereas, the command I shared were running one vllm replicas in each GPU and in reality reaching higher throughput with "data parallelsim" serving. Such as some answer [here ](https://github.com/vllm-project/vllm/discussions/294)

demo_apps/llama-on-prem.md
WuhanMonkey: So let me rephrase my original question, because it might caused some confusion here. The questions is: "If we pip install the vllm packages already, can we launch vllm api servers?". I think the answer is yes since I just tried on my VM. That said the alternative way is to clone the source repo, build from there (as well as install the dependencies), cd into the folder and run the scripts there. You shouldn't need to do both, especially we shouldn't rely on using package of dependencies installed from pip to build/run the source code because there could be a dependency mis-match.

demo_apps/llama-on-prem.md
```diff
@@ -0,0 +1,184 @@
...
+    }'
+```
+
+Also, if you have multiple GPUs, you can add the `--tensor-parallel-size` argument when starting the server (see [here](https://vllm.readthedocs.io/en/latest/serving/distributed_serving.html) for more info). For example, the command below runs the Llama 2 13b-chat model on 4 GPUs:
+
+```
+python api_server.py --host 0.0.0.0 --port 5000 --model meta-llama/Llama-2-13b-chat-hf --tensor-parallel-size 4
+```
+With multiple GPUs, you can also run replica of models as long as your model size can fit into targeted GPU memory. For example, if you have two A10G with 24 GB memory, you can run two 7B Llama 2 models at the same time. This can be done by launching two api servers each targeting specific CUDA cores on different ports:
+`CUDA_VISIBLE_DEVICES=0 python api_server.py --host 0.0.0.0 --port 5000  --model meta-llama/Llama-2-7b-chat-hf --tensor-parallel-size 1 `
```
HamidShojanazeri, line 67: @jeffxtang I am not sure if we need `--tensor-parallel-size 1 ` here as thats the default.

demo_apps/llama-on-prem.md
```diff
@@ -0,0 +1,184 @@
...
+and
+`CUDA_VISIBLE_DEVICES=1 python api_server.py --host 0.0.0.0 --port 5001  --model meta-llama/Llama-2-7b-chat-hf --tensor-parallel-size 1 `
+The benefit would be now you can balance incoming requests to both models, reaching higher batch size processing for a trade-off of generation latency.
+
```
HamidShojanazeri, line 71: It would be helpful to add that in case of serving multiple replica of the model using a serving solutions that has integrated vLLM into them make the load balancer available and add necessary metric and logs.

HamidShojanazeri (commented):
Thanks Jeff, added few minor comment, overall I wonder if that make sense to add either vllm or tgi to one of your existing notebook to kind of showcase the e2e story?

demo_apps/llama-on-prem.md
```diff
@@ -0,0 +1,184 @@
...
+
+### Deploying Llama 2 as OpenAI-Compatible Server
+
+You can also deploy the vLLM hosted Llama 2 as an OpenAI-Compatible service to easily replace code using OpenAI API. First, run the command below:
```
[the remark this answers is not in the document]
jeffxtang: @HamidShojanazeri the "why" was briefly started as "to easily replace code using OpenAI API" so people who are using  OpenAI (most of the LangChain and LlamaIndex examples and most blogs on building LLM apps use OpenAI) and considering switching to Llama2 can minimize their code change, hence called "OpenAI-Compatible".

demo_apps/llama-on-prem.md
jeffxtang: @HamidShojanazeri actually @WuhanMonkey Hector added this paragraph and looks like you're right - with CUDA_VISIBLE_DEVICES set to be a single value, --tensor-parallel-size shouldn't be needed. Hector can you please confirm?

demo_apps/llama-on-prem.md
jeffxtang: @WuhanMonkey Chester please comment/add?

demo_apps/llama-on-prem.md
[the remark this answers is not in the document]
jeffxtang: And the example code is in the Colab linked in the PR.

demo_apps/llama-on-prem.md
WuhanMonkey: True. Let me change it

demo_apps/llama-on-prem.md
WuhanMonkey: Sounds good. Let's add this context

demo_apps/llama-on-prem.md
WuhanMonkey, line 67:
```suggestion
`CUDA_VISIBLE_DEVICES=0 python api_server.py --host 0.0.0.0 --port 5000  --model meta-llama/Llama-2-7b-chat-hf`
```

demo_apps/llama-on-prem.md
WuhanMonkey: @HamidShojanazeri For you suggestion, do you actually mean we should provide a serving solution as an example that integrated vLLM with load balancer? Or do you mean we should jus tell them to use a serving solution with load balancer and metrics/load?

HamidShojanazeri: we both, we have an example WIP on Torchserve for this, but overall pointing out help to draw a more clear picture.

WuhanMonkey: OK, I do not have a solid example that built-in these serving solution already. I wrote the balancer with simple round-robin myself. My intention on this paragraph is to share that besides tensor parallelism, using model replicas could also improve serving performance. If a solid example is a required, maybe we can take the paragraph out for now. Really don't want to block on landing this PR.

HamidShojanazeri: No, I didn't mean to have a solid example, I was just mentioning to add this text thats it, never mind though. Lets merrge it as it user would be able to infer I guess.

HamidShojanazeri (approved):

PR #280, opened 2023-11-01
Add gradient clipping feature

### LINKED ISSUE #277: Is gradient clipping supported?
### 🚀 The feature, motivation and pitch

Does `llama-recipes` support `gradient clipping`?

I added an additional loss term to the original cross entropy loss and found that gradients (sometimes) become `NaN` after some iterations.

I could help add this feature if needed.

### Alternatives

_No response_

### Additional context

_No response_

gpengzhi:
# What does this PR do?

Fixes #277 

- [ ] Test A
Logs for Test A

- [ ] Test B
Logs for Test B

## Before submitting
      Pull Request section?
      to it if that's the case.

Thanks for contributing 🎉!


gpengzhi:
@HamidShojanazeri Could you help review this PR?

shreydan:
hello @gpengzhi I was going through the code for the changes in the PR. I think for gradient clipping we need to unscale the grads first if using mixed-precision. Since gradient-accumulation is also used, we've to perform the clipping, just before we step the optimizer.

example from torch docs: [amp_examples](https://pytorch.org/docs/stable/notes/amp_examples.html#gradient-clipping)

```py
scaler = GradScaler()

for epoch in epochs:
    for i, (input, target) in enumerate(data):
        with autocast(device_type='cuda', dtype=torch.float16):
            output = model(input)
            loss = loss_fn(output, target)
            loss = loss / iters_to_accumulate

        # Accumulates scaled gradients.
        scaler.scale(loss).backward()

        if (i + 1) % iters_to_accumulate == 0:
            # may unscale_ here if desired (e.g., to allow clipping unscaled gradients)
	     # Unscales the gradients of optimizer's assigned params in-place
            scaler.unscale_(optimizer)
            # Since the gradients of optimizer's assigned params are unscaled, clips as usual:
            torch.nn.utils.clip_grad_norm_(model.parameters(), max_norm)

            scaler.step(optimizer)
            scaler.update()
            optimizer.zero_grad()
```

gpengzhi:
> ```python
> scaler.unscale_(optimizer)
> ```

Thanks for the helpful comment!

You are right. We should clip the gradient after the accumulation step. I will refactor the code accordingly.

src/llama_recipes/utils/train_utils.py
```diff
@@ -87,6 +87,9 @@ def train(model, train_dataloader,eval_dataloader, tokenizer, optimizer, lr_sche
                     # if fp16 is enabled, use gradient scaler to handle gradient update
                     scaler.scale(loss).backward()
                     if (step + 1) % gradient_accumulation_steps == 0 or step == len(train_dataloader) - 1:
+                        if train_config.gradient_clipping > 0.0:
+                            scaler.unscale_(optimizer)
+                            torch.nn.utils.clip_grad_norm_(model.parameters(), train_config.gradient_clipping)
```
HamidShojanazeri, line 92: @gpengzhi thanks for the PR, in case FSDP we would need to use its own [clip_grad](https://github.com/pytorch/pytorch/blob/main/torch/distributed/fsdp/fully_sharded_data_parallel.py#L1067) as its shard aware. It would be `model.clip_grad_norm_(clip_threshold)`. Can you pls modify accordingly to cover FSDP as well.

gpengzhi: Thank you for the helpful feedback! 

Refactored the code to take care of the FSDP scenario.

src/llama_recipes/configs/training.py
```diff
@@ -14,6 +14,7 @@ class train_config:
     batching_strategy: str="packing" #alternative: padding
     context_length: int=4096
     gradient_accumulation_steps: int=1
+    gradient_clipping: float = 1.0
```
HamidShojanazeri, line 17: @gpengzhi I wonder if we should handle the default in a way to avoid any corner case. Maybe adding a bool, gradient_clipping and gradient_clipping_threshold, then guard gradient_clipping_threshold behind  gradient_clipping to make sure we don't run in to any edge case.

gpengzhi: An additional `bool` parameter is added to decide whether to utilize gradient clipping during training.

HamidShojanazeri (approved):
Thanks @gpengzhi LGTM

gpengzhi:
It seems that the CI pipeline fails when checking links in all markdown files. Should we merge this PR directly or disable the checking step for some specific links (something like https://github.com/facebookresearch/llama-recipes/pull/306/commits/52197ed6890eba83b5b8a289611797d8423216ed) in this PR? @HamidShojanazeri

HamidShojanazeri:
> It seems that the CI pipeline fails when checking links in all markdown files. Should we merge this PR directly or disable the checking step for some specific links (something like [52197ed](https://github.com/facebookresearch/llama-recipes/commit/52197ed6890eba83b5b8a289611797d8423216ed)) in this PR? @HamidShojanazeri
@gpengzhi sorry for late reply, yes pls let disable it as shown in that commit.

gpengzhi:
> @gpengzhi it seems its still failing with same error, can you pls follow the same commit here, [7dc9a11](https://github.com/facebookresearch/llama-recipes/commit/7dc9a11a1f2d5e7e626327b3a355a4aaf7645d48), basically you need to update to main now.

Updated

HamidShojanazeri:
thanks @gpengzhi for your contribution!

PR #287, opened 2023-11-07
A tutorial of building a Llama-enabled WhatsApp Chatbot

jeffxtang:
# What does this PR do?

Tutorial here: https://github.com/jeffxtang/llama-recipes/blob/whatsapp/demo_apps/whatsapp_llama2.md

README's also updated.

Fixes # (issue)

- [ ] Test A
Logs for Test A

- [ ] Test B
Logs for Test B

## Before submitting
      Pull Request section?
      to it if that's the case.
- [ X] Did you make sure to update the documentation with your changes?  

Thanks for contributing 🎉!


demo_apps/whatsapp_llama2.md
```diff
@@ -0,0 +1,151 @@
...
+## Modifying the Webhook 
+
+Open your glitch.com webhook URL created earlier, and after the code snippet:
+
+```
+// message received! 
+console.log(req.body["entry"][0]["changes"][0]["value"]["messages"][0]["text"]["body"]);
+```
+
+add the code below - remember to change <your web server IP>:
```
albertodepaola, line 118: Before this step, the server IP needs to be publicly visible, right? Might be good to point this out in the guide, just to cover make users aware of the need for a public service running the app.

demo_apps/whatsapp_llama2.md
```diff
@@ -0,0 +1,151 @@
...
+        return response.status_code
+```
+
+Finally, add the code below to llama-chatbot.py, which creates a Llama 2 instance and defines an HTTP method `msgrcvd` to:
+1. receive the user message forwarded by the webhook;
+2. ask Llama 2 for the answer;
+3. call the `WhatsAppClient`'s `send_text_message`` with a recipient's phone number.
+   
+```   
+os.environ["REPLICATE_API_TOKEN"] = "<your replicate api token>"
```
albertodepaola, line 85: Same here, we can point out that a replicate account is needed, as obvious as it is.

demo_apps/whatsapp_llama2.md
```diff
@@ -0,0 +1,151 @@
...
+
+@app.route("/")
+def hello_llama():
+    return "<p>Hello Llama 2</p>"
+
+@app.route('/msgrcvd', methods=['POST', 'GET'])
+def msgrcvd():    
+    message = request.args.get('message')
+    answer = llm(message)
+    client.send_text_message(llm(message), "<a recipient phone number from your WhatsApp API Setup>")
```
albertodepaola, line 103: Shouldn't the answer be used here to send the message instead of calling llm(message) again?

demo_apps/whatsapp_llama2.md
```diff
@@ -0,0 +1,151 @@
...
+
+```
+// message received! 
+console.log(req.body["entry"][0]["changes"][0]["value"]["messages"][0]["text"]["body"]);
+```
+
+add the code below - remember to change <your web server IP>:
+
+```
+  let url = "http://<your web server IP>:5000/msgrcvd?message=" +
```
albertodepaola, line 121: This could come from .env variable as well.

demo_apps/whatsapp_llama2.md
jeffxtang: yes the server IP needs to be publicly visible, before or after this step (but before the last step "open your WhatsApp app, enter a question"). Updated based on your feedback.

demo_apps/whatsapp_llama2.md
jeffxtang: I mentioned Replicate and other options in the first paragraph under "Writing Llama 2 Enabled Web App" so think that should be good.

demo_apps/whatsapp_llama2.md
jeffxtang: absoutely!

demo_apps/whatsapp_llama2.md
jeffxtang: yes but i guess i'll just leave it as is..

demo_apps/whatsapp_llama2.md
```diff
@@ -0,0 +1,151 @@
...
+
+add the code below - remember to change <web server public IP>, which needs to be publicly visible, to the IP of the server where your Llama 2 enabled web app in the previous section runs:
+
+```
+  let url = "http://<web server public IP>:5000/msgrcvd?message=" + 
+    req.body["entry"][0]["changes"][0]["value"]["messages"][0]["text"]["body"]
+
+  axios.get(url)
+    .then(response => {
+      console.log(response.data);
```
albertodepaola, line 126: From this code, it's not clear that the response gets sent back to the user via WA. It might be good to add the full code here, for reference.

jeffxtang: Good point - but actually the following code the app.js should be deleted because "the response gets sent back to the user via WA" is now done in the llama_chatbot.py. I updated the tutorial on this.

albertodepaola:
This looks great to me! not sure how to solve the spellcheck thou, it's erroring on technical terms.

jeffxtang:
> This looks great to me! not sure how to solve the spellcheck thou, it's erroring on technical terms.

Thanks! 

The fix is to add those missing words to scripts/spellcheck_conf/wordlist.txt - i just did.

HamidShojanazeri (approved):
thanks @jeffxtang LGTM.

PR #325, opened 2023-12-12
Changing Llama Guard safety check to HF classes

albertodepaola:
# What does this PR do?
Changes the Llama Guard safety check to use HF classes. Removes the previously used code to load Llama Guard.

* Llama guard enabled:
* Command: `python examples/inference.py --model_name ../llama/models_hf/7B --prompt_file examples/test_prompt_2.txt --enable_llamaguard_content_safety --enable_salesforce_content_safety False`
* Result
image

* Llama Guard disabled:
* Command: `python examples/inference.py --model_name ../llama/models_hf/7B --prompt_file examples/test_prompt_2.txt`
* Result
image

* Salesforce, Llama Guard and sensitive topics enabled:
* Command: `python examples/inference.py --model_name ../llama/models_hf/7B --prompt_file examples/test_prompt_2.txt --enable_llamaguard_content_safety --enable_sensitive_topics`
* Result
image

## Before submitting
      Pull Request section?
      to it if that's the case.

Thanks for contributing 🎉!


src/llama_recipes/inference/safety_utils.py
```diff
@@ -92,7 +91,7 @@ def get_total_length(self, data):
 
 # Class for performing safety checks using Azure Content Safety service
 class AzureSaftyChecker(object):
-    def __init__(sel, **kwargs):
+    def __init__(sel):
```
mreso, line 94: Not part of this PR but self is mistyped as sel but then references as self

mreso (approved):
Marked a bug in the comments which was already there but we should still fix it, otherwise LGTM

HamidShojanazeri (commented):
@albertodepaola thanks for the PR, I wonder if we need a separate example? `llama_gurda_inference.py`. 

Also, I wonder if still are keeping the llama_guard folder? if so I suggest to move this example there as it seems we still having the prompt_formatter there.

Please also add a llama_guard announcement on the main REAMDE and add a link to the respective folder/example.

albertodepaola:
@HamidShojanazeri The reference to the Llama Guard scripts in the relevant Readmes! 
@QingHu1227 Referencing the prompt formatting methods in the stand alone script only. For the safety checker, I'm using the method provided by the tokenizer. We can evaluate adding custom categories to that as well in the future.

QingHu1227 (approved):
The prompt format change looks good to me.

PR #326, opened 2023-12-12
Create finetuning data formatter for llama-guard

MichaelTontchev:
This creates a utility for formatting custom data for finetuning of llama-guard

Tests:
image

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,11 @@
+{
+    "python.testing.unittestArgs": [
+        "-v",
+        "-s",
+        "./tests",
+        "-p",
+        "test_*.py"
+    ],
+    "python.testing.pytestEnabled": false,
+    "python.testing.unittestEnabled": true
+}
--- a/src/llama_recipes/data/llama_guard/finetuning_data_formatter.py
+++ b/src/llama_recipes/data/llama_guard/finetuning_data_formatter.py
@@ -91,7 +91,7 @@ def create_formatted_finetuning_examples(
     Some of these augmented transformations are made by duplicating training
     examples and safely removing some violation categories from the llama
     guard prompts. Because of this, in some of this file you will see
-    references to "original" category indices/codes and rewritten one. The originals
+    references to "original" category indices/codes and rewritten ones. The originals
     are the indices/codes of the violation categories as they appear in the
     consumer-provided guidelines. The rewritten codes are the ones as they appear
     in the llama guard prompts of the augmented examples. We occasionally need to
@@ -143,7 +143,7 @@ def _verify_formatter_configs(
 
     if (
         formatter_configs.augmentation_configs.should_add_examples_with_dropped_violated_and_nonviolated_prompt_categories
-        > 0
+        == True
         and formatter_configs.llama_guard_generation_configs.explanation_position
         is not None
         and formatter_configs.augmentation_configs.explanation_for_augmentation_with_dropped_violated_and_nonviolated_prompt_categories
@@ -185,10 +185,6 @@ def _create_formatted_finetuning_example(
     return f"{llama_guard_prompt} {llama_guard_generation}"
 
 
-def _is_a_prompt_only_example(training_example: TrainingExample) -> bool:
-    return training_example.response == "N/A"
-
-
 def _create_llama_guard_prompt(
     training_example: TrainingExample,
     category_indices_to_include: List[int],
@@ -221,6 +217,10 @@ def _create_llama_guard_prompt(
             "conversation": _serialize_conversation(conversation),
         }
     )
+
+
+def _is_a_prompt_only_example(training_example: TrainingExample) -> bool:
+    return training_example.response == "N/A"
 
 
 def _serialize_conversation(conversation: Dict[str, str]) -> str:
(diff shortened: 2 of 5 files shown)


src/llama_recipes/data/llama_guard/finetuning_data_formatter.py
```diff
@@ -0,0 +1,481 @@
...
+        to_return = f"Explanation: {training_example.explanation}\n{to_return}"
+    elif explanation_position == ExplanationPosition.AFTER_DECISION:
+        to_return = f"{to_return}\nExplanation: {training_example.explanation}"
+
+    return to_return
+
+
+def _get_map_of_original_category_indices_to_rewritten_category_codes(
+    formatter_configs: FormatterConfigs,
+    category_indices_included_in_llama_guard_prompt: List[int],
```
inanhkn, line 281: could we order the indices so that the output lists them in ascending order?

inanhkn (commented):
Looks very comprehensive! added some comments.

.vscode/settings.json
```diff
@@ -0,0 +1,11 @@
+{
```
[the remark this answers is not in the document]
MichaelTontchev: If we don't have it, test discovery isn't set up for users in codespaces. Imo promotes testing

src/llama_recipes/data/llama_guard/finetuning_data_formatter.py
```diff
@@ -0,0 +1,481 @@
...
+    return to_return
+
+
+def _verify_formatter_configs(
+    formatter_configs: FormatterConfigs,
+) -> None:
+    if (
+        formatter_configs.augmentation_configs.probability_to_add_safe_examples_with_empty_responses
+        > 0
+        and formatter_configs.llama_guard_generation_configs.explanation_position
```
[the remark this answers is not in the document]
MichaelTontchev: I think it does already? If explanation position is none, an explanation is not required. I need to update the type on the explanation, but otherwise, it's already there, no?

src/llama_recipes/data/llama_guard/finetuning_data_formatter.py
MichaelTontchev: Is there a reason for this? Does the ordering have meaning in the output?

.vscode/settings.json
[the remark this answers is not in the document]
HamidShojanazeri: @MichaelTontchev can you pls share a bit more what are we referring to with codespaces here?

HamidShojanazeri (commented):
thanks @MichaelTontchev for the PR, it would be great if could add a readme here, describing what is this data formatter doing, why people should use it, and example of usage with one of the datasets ideally ( or maybe in the next PR).

Also, wonder if this would be the same for HF checkpoints as well.

src/llama_recipes/data/llama_guard/finetuning_data_formatter.py
inanhkn: it's deterministic behavior, which is desired over non-deterministic behavior

src/llama_recipes/data/llama_guard/finetuning_data_formatter.py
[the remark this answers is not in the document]
inanhkn: Yes, I just wanted to point out the main use case without explanations. As long as we make sure and add examples in unit tests without explanations, it's perfect.

src/llama_recipes/data/llama_guard/finetuning_data_formatter.py
```diff
@@ -0,0 +1,481 @@
...
+
+
+@dataclass
+class Category:
+    name: str
+    description: str
+
+
+@dataclass
+class Guidelines:
```
albertodepaola, line 18: @MichaelTontchev This is a nice abstraction for handling prompt categories, maybe we can extract it and put it in the [prompt_format_utils](https://github.com/facebookresearch/llama-recipes/blob/main/src/llama_recipes/inference/prompt_format_utils.py#L14) in place of the SafetyCategories.

albertodepaola: cc @QingHu1227

src/llama_recipes/data/llama_guard/finetuning_data_formatter.py
```diff
@@ -0,0 +1,481 @@
...
+        formatter_configs,
+    )
+
+    llama_guard_generation = _create_llama_guard_generation(
+        training_example,
+        formatter_configs,
+        category_indeces_to_include_in_llama_guard_prompt,
+    )
+
+    return f"{llama_guard_prompt} {llama_guard_generation}"
```
albertodepaola, line 185: @MichaelTontchev I noticed in the generated output that no new line is present between the end of the input prompt and the generation response. For inference, they are separated by a new line, should a `\n` be included here between the prompt and the generation as well?

src/llama_recipes/data/llama_guard/finetuning_data_formatter.py
QingHu1227: Sounds a good idea to me:)

albertodepaola (approved):
Code works as expected, some comments and Readme will be addressed in future PR.

src/llama_recipes/data/llama_guard/finetuning_data_formatter.py
albertodepaola: Will come in a later PR!

albertodepaola (approved):
Retested with the script and unittests locally and everything worked correctly.

PR #342, opened 2023-12-25
Update Installation section of README.md

daniellepintz:
* Move `Install with optional dependencies` section above `Install from source` since it should go next to the section about installing with pip
* Add `git clone` and `cd llama-recipes` to `Install from source` section


mreso (approved):
LGTM

PR #365, opened 2024-02-05
Codellama 70b instruct code example

HamidShojanazeri:
# What does this PR do?
Adding code example for prompting codellama 70B instruct. model

- [ ] Test A
[Logs for e2e test](https://gist.github.com/HamidShojanazeri/f304527212cc52cb93846126698353dd)

- [ ] Test B
Logs for Test B

## Before submitting
      Pull Request section?
- [ x] Was this discussed/approved via a Github issue? Please add a link
      to it if that's the case.
- [ x] Did you make sure to update the documentation with your changes?  

Thanks for contributing 🎉!

### DIFF
--- a/docs/inference.md
+++ b/docs/inference.md
@@ -86,7 +86,7 @@ To run the 70B Instruct model example run the following, it asks for system and 
 python code_instruct_example.py --model_name codellama/CodeLlama-70b-Instruct-hf
 
 ```
-You can learn more about the chat prompt template [on HF](https://huggingface.co/codellama/CodeLlama-70b-Instruct-hf#chat-prompt) and [original CodeLlama repository](https://github.com/facebookresearch/codellama/blob/main/README.md#fine-tuned-instruction-models). HF tokenizer has already taken care of the chat template as shown in this example. 
+You can learn more about the chat prompt template [on HF](https://huggingface.co/codellama/CodeLlama-70b-Instruct-hf#chat-prompt) and [original Code Llama repository](https://github.com/facebookresearch/codellama/blob/main/README.md#fine-tuned-instruction-models). HF tokenizer has already taken care of the chat template as shown in this example. 
 
 ### Llama Guard
 
--- a/docs/inference.md
+++ b/docs/inference.md
@@ -79,7 +79,13 @@ To run the code infilling example:
 python examples/code_llama/code_infilling_example.py --model_name MODEL_NAME --prompt_file examples/code_llama/code_infilling_prompt.txt --temperature 0.2 --top_p 0.9
 
 ```
+To run the 70B Instruct model example run the following, it asks for system and user prompt to instruct the model:
 
+```bash
+
+python code_instruct_example.py --model_name codellama/CodeLlama-70b-Instruct-hf
+
+```
 ### Llama Guard
 
 Llama Guard is a new experimental model that provides input and output guardrails for LLM deployments. For more details, please visit the main [repository](https://github.com/facebookresearch/PurpleLlama/tree/main/Llama-Guard).
--- a/docs/inference.md
+++ b/docs/inference.md
@@ -86,6 +86,8 @@ To run the 70B Instruct model example run the following, it asks for system and 
 python code_instruct_example.py --model_name codellama/CodeLlama-70b-Instruct-hf
 
 ```
+You can learn more about the chat prompt template [on HF](https://huggingface.co/codellama/CodeLlama-70b-Instruct-hf#chat-prompt) and [original CodeLlama repository](https://github.com/facebookresearch/codellama/blob/main/README.md#fine-tuned-instruction-models). HF tokenizer has already taken care of the chat template as shown in this example. 
+
 ### Llama Guard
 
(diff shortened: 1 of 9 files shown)


examples/code_llama/code_completion_example.py
```diff
@@ -33,6 +33,7 @@ def main(
     enable_azure_content_safety: bool=False, # Enable safety check with Azure content safety api
     enable_sensitive_topics: bool=False, # Enable check for sensitive topics using AuditNLG APIs
     enable_salesforce_content_safety: bool=True, # Enable safety check with Salesforce safety flan t5
+    enable_llamaguard_content_safety: bool=False, # Enable safety check with Llama-Guard
```
varunfb, line 36: Looking at the code path examples/code_llama
Do we want to have Llama Guard model for Code Models? CyberSecEvals is a great benchmark to use post fine tuning of any code models and Llama Guard for content moderation.

examples/code_llama/code_infilling_example.py
```diff
@@ -32,6 +32,7 @@ def main(
     enable_azure_content_safety: bool=False, # Enable safety check with Azure content safety api
     enable_sensitive_topics: bool=False, # Enable check for sensitive topics using AuditNLG APIs
     enable_salesforce_content_safety: bool=True, # Enable safety check with Salesforce safety flan t5
+    enable_llamaguard_content_safety: bool=False, # Enable safety check with Llama-Guard
```
varunfb, line 35: Same comment as the previous one. Do we need Llama Guard for code models?

examples/code_llama/code_completion_example.py
HamidShojanazeri: We had to add it here as it has been added to the [`get_saferty_checker`](https://github.com/facebookresearch/llama-recipes/blob/main/src/llama_recipes/inference/safety_utils.py#L209) util, so kind of had to add it to make the code functional. But as this is generally a safety check on the prompts, I kind of think its pretty much the same with other safety packages we have in place already, so that might be just an addition to the mix.

examples/code_llama/code_infilling_example.py
HamidShojanazeri: I just replied above, also to add that people can choose not to run the safety check at all in these scripts.

docs/inference.md
```diff
@@ -79,6 +79,14 @@ To run the code infilling example:
 python examples/code_llama/code_infilling_example.py --model_name MODEL_NAME --prompt_file examples/code_llama/code_infilling_prompt.txt --temperature 0.2 --top_p 0.9
 
 ```
+To run the 70B Instruct model example run the following, it asks for system and user prompt to instruct the model:
+
+```bash
+
+python code_instruct_example.py --model_name codellama/CodeLlama-70b-Instruct-hf
```
jeffxtang, line 86: add `examples/code_llama/` before code_instruct_example.py ?

also , should we mention that it'd take about 10 minutes and 130GB free space to download all the shard safetensors model files?

docs/inference.md
```diff
@@ -79,6 +79,14 @@ To run the code infilling example:
 python examples/code_llama/code_infilling_example.py --model_name MODEL_NAME --prompt_file examples/code_llama/code_infilling_prompt.txt --temperature 0.2 --top_p 0.9
 
 ```
+To run the 70B Instruct model example run the following, it asks for system and user prompt to instruct the model:
```
jeffxtang, line 82: => To run the 70B Instruct model example (you'll need to enter the system and user prompts to instruct the model):

docs/inference.md
jeffxtang: I got the following error when running the command above (as well as `python examples/code_llama/code_completion_example.py --model_name codellama/CodeLlama-7b-Instruct-hf --prompt_file examples/code_llama/code_completion_prompt.txt --temperature 0.2 --top_p 0.9`) after all the model files were downloaded (notice my transformers and torch versions are shown to be ok and don't need to be upgraded):
image

This is how I installed the packages inside a new conda environment with python 3.11.7:
```
pip install git+https://github.com/huggingface/transformers
pip install --extra-index-url https://download.pytorch.org/whl/test/cu118 llama-recipes
```

examples/code_llama/code_instruct_example.py
```diff
@@ -0,0 +1,158 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# This software may be used and distributed according to the terms of the Llama 2 Community License Agreement.
+
+# from accelerate import init_empty_weights, load_checkpoint_and_dispatch
```
jeffxtang, line 4: remove the comment?

jeffxtang (changes requested):
@HamidShojanazeri example looks great overall. just a few comments (and one error preventing me from actually running it)

docs/inference.md
HamidShojanazeri: sounds like an [issue](https://github.com/huggingface/transformers/issues/28877) on HF side. we can disable the` --fast_kernel False` for now until the fix come along.

HamidShojanazeri: This has been resolved with `attn_implementation="eager`, all references to better_transformer is now updated.

jeffxtang (approved):

PR #390, opened 2024-03-11
removing legacy code for sdpa

HamidShojanazeri:
# What does this PR do?

Removing the legacy code as we integrated new way HF adapted for using SDPA such as other [examples](https://github.com/facebookresearch/llama-recipes/blob/main/examples/inference.py#L73), this example was not updated.


mreso (approved):
LGTM

PR #398, opened 2024-03-14
Add llm class so that externally-hosted models can be called

tryrobbo:
This change adds the ability  to call out to 3rd party hosted models on OpenAI, Together AI and Anyscale

Test case:
```
from llama_recipes.inference.llm import  TOGETHER, OPENAI, ANYSCALE

together_example = TOGETHER("togethercomputer/llama-2-7b-chat","09e45...")
print( together_example.query(prompt="Why is the sky blue?"))


openai_example = OPENAI("gpt-3.5-turbo","sk-LIz9zL3cYp...")
print( openai_example.query(prompt="Why is the sky blue?"))


anyscale_example = ANYSCALE("meta-llama/Llama-2-7b-chat-hf","esecret_c3u4x7...")
print( anyscale_example.query(prompt="Why is the sky blue?"))
```


src/llama_recipes/inference/llm.py
```diff
@@ -0,0 +1,228 @@
...
+        return self._query_with_retries(
+            self.query_with_system_prompt, system_prompt, prompt
+        )
+
+    def valid_models(self) -> list[str]:
+        """List of valid model parameters, e.g. 'gpt-3.5-turbo' for GPT"""
+        return []
+
+
+SPECIFICATION_FORMAT: str = "<PROVIDER>::<MODEL>(::<API KEY>)?"
```
varunfb, line 101: This was specific to cyberseceval. Do we need this?

src/llama_recipes/inference/llm.py
```diff
@@ -0,0 +1,228 @@
...
+EXAMPLE_SPECIFICATION: str = "OPENAI::gpt-3.5-turbo::<YOUR API KEY>"
```
varunfb, line 102: And this !!

tryrobbo:
Hi @HamidShojanazeri , would you mind taking a look at this please?

src/llama_recipes/inference/llm.py
```diff
@@ -0,0 +1,197 @@
...
+        return self._query_with_retries(
+            self.query_with_system_prompt, system_prompt, prompt
+        )
+
+    def valid_models(self) -> list[str]:
+        """List of valid model parameters, e.g. 'gpt-3.5-turbo' for GPT"""
+        return []
+
+
+class OPENAI(LLM):
```
HamidShojanazeri, line 101: Also, wondering if we need to keep the specific APIs out of this module and keep them as examples? otherwise it means one need to install all these packages/ adding all as dependency? openai, ocotoAI etc?

HamidShojanazeri:
Thanks @tryrobbo for the great PR, just left some minor comments. Also it would be great to add documentations/ examples.

src/llama_recipes/inference/llm.py
```diff
@@ -0,0 +1,197 @@
...
+    """Accessing OPENAI"""
+
+    def __init__(self, model: str, api_key: str) -> None:
+        super().__init__(model, api_key)
+        self.client = openai.OpenAI(api_key=api_key)  # noqa
+
+    @override
+    def query(self, prompt: str) -> str:
```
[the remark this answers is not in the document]
tryrobbo: At the moment the intention is to provide an example that is as simple as possible. I agree that retries would be a good enhancement though, and definitely best practice.

src/llama_recipes/inference/llm.py
tryrobbo: I have deliberately left these modules out of the requirements.txt to avoid having a module level dependency on these modules. I think that users can install them selectively as they need to use them and get an error when they are not found. Do you think that's a reasonable middle ground?

tryrobbo:
Many thanks for your thoughts, @HamidShojanazeri . I've added a notebook with a brief explanation and examples - I think the code broadly speaks for itself. Please let me know if you think this is in a reasonable state to merge now. Thanks!

HamidShojanazeri (approved):
Thanks @tryrobbo, let merge it to unblock, let take the other improvements in the follow up PRs.

PR #485, opened 2024-05-03
Add LangChain recipes

rlancemartin:
# What does this PR do?

Add recipes for LangChain agents.

Notebooks tested in fresh conda env.


jeffxtang (commented):
@rlancemartin 
Thanks for the PR Lance! I made and pushed some changes (summary in the commit message) to the 4 notebooks and will continue to play with it later. I noticed in the tool-calling-agent notebook, using `app.invoke({"question":"What is LangSmith?"})` would return correct result. In the local notebook, the answer to "What are the types of agent memory?" is not as good as the non-local Groq's llama 3 70b based one. 

Please take a look at my changes and let me know what you think. I'll see if it's possible to simplify langgraph-agent to make the benefits of using LG over AE obvious.

jeffxtang (approved):

PR #487, opened 2024-05-05
Fix param_init_fn: move if-statement out of lambda

pbelevich:
# What does this PR do?

Without parentheses if-statement becomes a part of lambda which is incorrect.

cc @lchu-ibm @KeitaW


HamidShojanazeri (approved):
Thanks @pbelevich for the PR!

PR #507, opened 2024-05-12
Fixing path issues

tokenizer-decode:
Preventing errors in case the directory doesn't exist.


src/llama_recipes/utils/config_utils.py
tokenizer-decode, line 1: I also think we shouldn't warn but assert here

mreso: Thanks for the suggestion, we will consider raising a ValueError here.

mreso (approved):
LGTM, thanks for the contribution

PR #516, opened 2024-05-15
Freeze layer bug fix

tokenizer-decode:
Fix bug when not using PEFT


mreso (approved):
LGTM, thanks for contributing!

PR #608, opened 2024-07-24
Add PromptGuard to safety_utils

tryrobbo:
[PromptGuard](https://github.com/meta-llama/PurpleLlama/tree/main/Prompt-Guard) has been introduced as a system safety tool to be used to check LLM prompts for malicious text. This PR adds this guard to the safety_utils module and adjusts recipes to use this new check. 

- Implementation details
  - PromptGuard has a limit of 512 tokens. Naive sentence splitting is performed on inputs to try to ensure that this limit is not breached. A warning is printed if a sentence is longer than 512 tokens. 
  - Only the jailbreak score of PromptGuard is acted on. The other scores are intended to be used in agentic implemetations, of which there are none currently which use the safety_utils module. 

General simple test
```
echo "hello" |python3 inference.py "meta-llama/Meta-Llama-3.1-8B-Instruct"  --quantization '8bit' --use_fast_kernels --enable-promptguard-safety True 
```

To test the text splitter feature
```
python3 inference.py "meta-llama/Meta-Llama-3.1-8B-Instruct"  --quantization '8bit' --use_fast_kernels --enable-promptguard-safety True < ~/longprompt 
```

### DIFF
--- a/recipes/quickstart/inference/code_llama/code_completion_example.py
+++ b/recipes/quickstart/inference/code_llama/code_completion_example.py
@@ -34,6 +34,7 @@ def main(
     enable_sensitive_topics: bool=False, # Enable check for sensitive topics using AuditNLG APIs
     enable_salesforce_content_safety: bool=True, # Enable safety check with Salesforce safety flan t5
     enable_llamaguard_content_safety: bool=False, # Enable safety check with Llama-Guard
+    enable_promptguard_safety: bool = False,
     use_fast_kernels: bool = True, # Enable using SDPA from PyTroch Accelerated Transformers, make use Flash Attention and Xformer memory-efficient kernels
     **kwargs
 ):
@@ -62,6 +63,7 @@ def main(
                                         enable_sensitive_topics,
                                         enable_salesforce_content_safety,
                                         enable_llamaguard_content_safety,
+                                        enable_promptguard_safety,
                                         )
 
     # Safety check of the user prompt
--- a/recipes/quickstart/inference/local_inference/chat_completion/chat_completion.py
+++ b/recipes/quickstart/inference/local_inference/chat_completion/chat_completion.py
@@ -37,6 +37,7 @@ def main(
     enable_saleforce_content_safety: bool=True, # Enable safety check woth Saleforce safety flan t5
     use_fast_kernels: bool = False, # Enable using SDPA from PyTorch Accelerated Transformers, make use Flash Attention and Xformer memory-efficient kernels
     enable_llamaguard_content_safety: bool = False,
+    enable_promptguard_safety: bool = False,
     **kwargs
 ):
     if prompt_file is not None:
@@ -81,6 +82,7 @@ def main(
                                         enable_sensitive_topics,
                                         enable_saleforce_content_safety,
                                         enable_llamaguard_content_safety,
+                                        enable_promptguard_safety
                                         )
             # Safety check of the user prompt
             safety_results = [check(dialogs[idx][0]["content"]) for check in safety_checker]
--- a/src/llama_recipes/inference/safety_utils.py
+++ b/src/llama_recipes/inference/safety_utils.py
@@ -205,34 +205,51 @@ class LlamaGuardSafetyChecker(object):
 class PromptGuardSafetyChecker(object):
 
     def __init__(self):
-        from transformers import AutoModelForCausalLM, AutoTokenizer, BitsAndBytesConfig
-        model_id = "llhf/Prompt-Guard-86M"
+        from transformers import AutoModelForSequenceClassification, AutoTokenizer, BitsAndBytesConfig
+        model_id = "meta-llama/Prompt-Guard-86M"
 
         self.tokenizer = AutoTokenizer.from_pretrained(model_id)
-        self.model = AutoModelForCausalLM.from_pretrained(model_id, device_map="auto")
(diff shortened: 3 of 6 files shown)


recipes/quickstart/inference/code_llama/code_completion_example.py
```diff
@@ -34,6 +34,7 @@ def main(
     enable_sensitive_topics: bool=False, # Enable check for sensitive topics using AuditNLG APIs
     enable_salesforce_content_safety: bool=True, # Enable safety check with Salesforce safety flan t5
     enable_llamaguard_content_safety: bool=False, # Enable safety check with Llama-Guard
+    enable_promptguard_safety: bool = False,
```
albertodepaola, line 37: Given the small size of the mode, can we leave it as default true?

recipes/quickstart/inference/local_inference/chat_completion/chat_completion.py
```diff
@@ -37,6 +37,7 @@ def main(
     enable_saleforce_content_safety: bool=True, # Enable safety check woth Saleforce safety flan t5
     use_fast_kernels: bool = False, # Enable using SDPA from PyTorch Accelerated Transformers, make use Flash Attention and Xformer memory-efficient kernels
     enable_llamaguard_content_safety: bool = False,
+    enable_promptguard_safety: bool = False,
```
albertodepaola, line 40: specially for this call, can we set it to true?

src/llama_recipes/inference/safety_utils.py
```diff
@@ -201,14 +201,55 @@ def __call__(self, output_text, **kwargs):
...
+        agent_type = kwargs.get('agent_type', AgentType.USER)
+        if agent_type == AgentType.AGENT:
+            return "PromptGuard", True, "PromptGuard is not used for model output so checking not carried out"
+        sentences = text_for_check.split(".")
+        running_scores = {'jailbreak':0, 'indirect_injection' :0}
+        for sentence in sentences:
+            scores = self.get_scores(sentence)
+            running_scores['jailbreak'] = max([running_scores['jailbreak'],scores['jailbreak']])
+            running_scores['indirect_injection'] = max([running_scores['indirect_injection'],scores['indirect_injection']])
+        is_safe = True if running_scores['jailbreak'] < 0.5 else False
```
albertodepaola, line 242: Should we set the bar at 0.5? I think 0.8 or 0.9 would be better based on talks with the team.

src/llama_recipes/inference/safety_utils.py
```diff
@@ -201,14 +201,55 @@ def __call__(self, output_text, **kwargs):
...
+    def __call__(self, text_for_check, **kwargs):     
+        agent_type = kwargs.get('agent_type', AgentType.USER)
+        if agent_type == AgentType.AGENT:
+            return "PromptGuard", True, "PromptGuard is not used for model output so checking not carried out"
+        sentences = text_for_check.split(".")
+        running_scores = {'jailbreak':0, 'indirect_injection' :0}
+        for sentence in sentences:
+            scores = self.get_scores(sentence)
+            running_scores['jailbreak'] = max([running_scores['jailbreak'],scores['jailbreak']])
+            running_scores['indirect_injection'] = max([running_scores['indirect_injection'],scores['indirect_injection']])
```
albertodepaola, line 241: Should we comment that this is not being used for the user dialog?

albertodepaola (changes requested):
Overall looks good! added a couple of simple comments only.

src/llama_recipes/inference/safety_utils.py
```diff
@@ -201,14 +201,55 @@ def __call__(self, output_text, **kwargs):
...
+        model_id = "meta-llama/Prompt-Guard-86M"
+
+        self.tokenizer = AutoTokenizer.from_pretrained(model_id)
+        self.model = AutoModelForSequenceClassification.from_pretrained(model_id)
+
+    def get_scores(self, text, temperature=1.0, device='cpu'):
+        from torch.nn.functional import softmax
+        inputs = self.tokenizer(text, return_tensors="pt", padding=True, truncation=True, max_length=512)
+        inputs = inputs.to(device)
+        if len(inputs[0]) > 512:
```
mreso, line 218: As max_length is 512 this condition should never be true. Can we instead follow the PromptGuard recommendation and split the text into multiple segments which we apply in parallel (batched)? Especially because of the much bigger context length of Llama 3.1.

src/llama_recipes/inference/safety_utils.py
```diff
@@ -201,14 +201,55 @@ def __call__(self, output_text, **kwargs):
...
+            'indirect_injection': (probabilities[0, 1] + probabilities[0, 2]).item()
+        }
+
+    def __call__(self, text_for_check, **kwargs):     
+        agent_type = kwargs.get('agent_type', AgentType.USER)
+        if agent_type == AgentType.AGENT:
+            return "PromptGuard", True, "PromptGuard is not used for model output so checking not carried out"
+        sentences = text_for_check.split(".")
+        running_scores = {'jailbreak':0, 'indirect_injection' :0}
+        for sentence in sentences:
```
mreso, line 238: Its probably more efficient to do this batched as commented above. Lets split the prompt in blocks of 512 (with some overlap) and then feed them batched into the model which will be way more efficient than feeding the sentences one by one.

mreso (changes requested):
Looks good overall but lets address the way we feed the prompt to the model (See comments).

PR #636, opened 2024-08-17
Add preprocessor to patch PromptGuard scores for inserted characters

cynikolai:
**Problem**: Inserting spaces between characters in given prompts causes misclassifications in PromptGuard. See https://github.com/meta-llama/llama-models/issues/50 for more context.

**Solution:** Tokenize the string with all spaces removed, to ensure that larger tokens (for example, `[“ignore”]`) are not broken up into smaller tokens (for example, `[“i”, “g”, “n”, “o”, “r”, “e”]` . Add back spaces between the larger tokens if spaces exist in the original string.

This approach showed a slight positive impact on all of our evaluation datasets, suggesting that making the system more robust to jailbreaks that disrupt tokenization like this one will be an important part of improving model quality. Notably, simply subtracting spaces from the string lead to a moderate quality regression on some datasets, which is why we don’t take that simpler approach here.

This solution only targets jailbreaks enabled by inserted spaces and not other special characters. For a more complete approach longer term, we’re continuing to work on building more adversarial examples into our dataset.

The preprocessor is used by default by our inference utilities.

### DIFF
--- a/recipes/responsible_ai/prompt_guard/inference.py
+++ b/recipes/responsible_ai/prompt_guard/inference.py
@@ -31,7 +31,45 @@ def load_model_and_tokenizer(model_name='meta-llama/Prompt-Guard-86M'):
     return model, tokenizer
 
 
-def get_class_probabilities(model, tokenizer, text, temperature=1.0, device='cpu'):
+def preprocess_text_for_promptguard(text: str, tokenizer) -> str:
+    """
+    Preprocess the text by removing spaces that break apart larger tokens.
+    This hotfixes a workaround to PromptGuard, where spaces can be inserted into a string
+    to allow the string to be classified as benign.
+
+    Args:
+        text (str): The input text to preprocess.
+        tokenizer (transformers.PreTrainedTokenizer): The tokenizer for the model.
+
+    Returns:
+        str: The preprocessed text.
+    """
+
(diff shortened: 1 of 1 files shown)


recipes/responsible_ai/prompt_guard/inference.py
```diff
@@ -31,7 +31,45 @@ def load_model_and_tokenizer(model_name='meta-llama/Prompt-Guard-86M'):
...
+        for i, char in enumerate(text):
+            if not char.isspace():
+                cleaned_text += char
+                index_map.append(i)
+        tokens = tokenizer.tokenize(cleaned_text)
+        result = []
+        last_end = 0
+        for token in tokens:
+            token_str = tokenizer.convert_tokens_to_string([token])
+            start = cleaned_text.index(token_str, last_end)
```
mreso, line 60: IIRC tokenizer are not always lossless. Could this be circumvented by adding a word/string combination into the text which can not be tokenized "losslessly" leading to a ValueError being raised in .index() as the "detokenized" string can not be found?

mreso (approved):
LGTM but this might be easily circumvented if the tokenizer is not lossless (see comment)

mreso:
Turns out tiktoken is actually lossless.

Issue #633, opened 2024-08-15
Clarification on Evaluation Results for Llama Guard 3


sheli-kohan:
### System Info

Versions of relevant libraries:
[pip3] mypy-extensions==0.4.3
[pip3] numpy==1.23.2
[pip3] torch==2.0.1+cu118
[pip3] torchaudio==2.0.2+cu118
[pip3] torchvision==0.15.2+cu118

### Information

- [X] The official example scripts
- [ ] My own modified scripts

### 🐛 Describe the bug

I am currently evaluating the Llama Guard 3 model using the evaluation notebook provided in the llama-recipes repo: [Llama Guard Customization via Prompting and Fine-Tuning](https://github.com/meta-llama/llama-recipes/blob/main/recipes/responsible_ai/llama_guard/llama_guard_customization_via_prompting_and_fine_tuning.ipynb).

When I ran the evaluation on the ToxicChat dataset, I observed an average precision of **30.20%**. This was with the following configurations: split="test".

However, I noticed a discrepancy when comparing this result to the [Llama Guard Model Card](https://github.com/meta-llama/PurpleLlama/blob/main/Llama-Guard/MODEL_CARD.md), which reports an average precision of **62.6%**.  even though the metric is referred to Llama Guard , I believe this degradation means some error in this notebook.

In other matter, we are failing to replicate paper results also for open ai mod eval dataset by category ([figure 2 in the paper](https://scontent.ftlv5-1.fna.fbcdn.net/v/t39.2365-6/408725049_3688557441468029_8103913771964668529_n.pdf?_nc_cat=100&ccb=1-7&_nc_sid=3c67a6&_nc_ohc=J9E3fv4CdawQ7kNvgGCWfkd&_nc_ht=scontent.ftlv5-1.fna&oh=00_AYBECymkNoF_oD98-bKT2VA4ckDSzjJToN8kNgrI4y_eWg&oe=66C3A759)). if you'll be able to share the library or code you used for this evaluation that will be very helpful

Could you please provide any insights or guidance on this difference in performance?

Thank you for your time and assistance.

Best regards,

Sheli Kohan

### Error logs

average precision 30.02%

### Expected behavior

average precision 62.6%

init27:
Thanks for flagging @sheli-kohan!

@albertodepaola Can you help take a look please?

sheli-kohan:
hi,

[...]

PR #653, opened 2024-08-29
fix that help reaching 50% over binary classification of toxic chat

sheli-kohan:
Initial fix for # https://github.com/meta-llama/llama-recipes/issues/633


sheli-kohan:
Closing — this has been stale since Aug 2024 and now conflicts with main. No longer pursuing.

PR #725, opened 2024-10-14
Fix fine-tuning training loss accumulation

celestinoalan:
# What does this PR do?

## Problem:
In /src/llama_recipes/utils/train_utils.py the training loss is correctly divided by the # of gradient accumulation steps to scale down the gradient:

loss = loss / gradient_accumulation_steps

The training loss is then accumulated

total_loss += loss.detach().float()

and used in the following to calculate the average loss across all samples in the epoch:

train_epoch_loss = total_loss / len(train_dataloader)

As the accumulated loss is scaled down by gradient_accumulation_steps and len(train_dataloader) includes all steps (even the gradient accumulation ones), train_epoch_loss is gradient_accumulation_steps times lower than it should be.

## Solution:

Accumulate the loss
total_loss += loss.detach().float()

before scaling it down

loss = loss / gradient_accumulation_steps

## Before submitting
- [ X ] Did you read the [contributor guideline](https://github.com/facebookresearch/llama-recipes/blob/main/CONTRIBUTING.md#pull-requests),
      Pull Request section?
      to it if that's the case.

Thanks for contributing 🎉!

### DIFF
--- a/src/llama_recipes/utils/train_utils.py
+++ b/src/llama_recipes/utils/train_utils.py
@@ -151,11 +151,11 @@ def train(model, train_dataloader,eval_dataloader, tokenizer, optimizer, lr_sche
                                 batch[key] = batch[key].to('cuda:0')
                     with autocast():
                         loss = model(**batch).loss
+                    total_loss += loss.detach().float()
                     loss = loss / gradient_accumulation_steps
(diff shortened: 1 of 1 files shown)


mreso (approved):
LGTM, thanks for contributing!

PR #744, opened 2024-10-24
Append epoch rather than best val. loss to val_loss

celestinoalan:
# Problem
Currently, we're val_loss.append(best_val_loss) in each epoch. This is misleading because we're appending the corresponding epoch (not best across epochs) quantities in train_loss, train_prep, and val_prep. This is also inconvenient, as one often would like to plot both train and validation losses as a function of the epochs to look for overfitting.

# Solution
val_loss.append(eval_epoch_loss)

## Before submitting
- [ X ] Did you read the [contributor guideline](https://github.com/facebookresearch/llama-recipes/blob/main/CONTRIBUTING.md#pull-requests),
      Pull Request section?
      to it if that's the case.

Thanks for contributing 🎉!


init27 (approved):

init27:
Thanks for the great catch-merged!

PR #775, opened 2024-11-06
Update samsum_dataset.py

iankouls-aws:
# What does this PR do?

add trust_remote_code=True to correctly load samsum dataset

Fixes error stating that samsum dataset contains custom code that needs to be trusted in order to properly load the data.

### DIFF
--- a/src/llama_recipes/datasets/samsum_dataset.py
+++ b/src/llama_recipes/datasets/samsum_dataset.py
@@ -8,7 +8,7 @@ import datasets
 
 
 def get_preprocessed_samsum(dataset_config, tokenizer, split):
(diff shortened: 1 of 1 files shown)


mreso:
Hi @iankouls-aws which version of llama-recipes are you using? [In main](https://github.com/meta-llama/llama-recipes/blob/b9fc10696ca1cac6536a9d16ec16616387ec7e43/src/llama_recipes/datasets/samsum_dataset.py#L17C203-L17C232) we prompt the user to set HF_DATASETS_TRUST_REMOTE_CODE variable to avoid the error. Silently enabling the execution of remote code could have security risks.

PR #838, opened 2025-01-12
Fix a property name of list_emails function

snowsky:
# What does this PR do?

Property name `maxResults` should be renamed as `max_results`.

## Before submitting
- [x] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).
      Pull Request section?
      to it if that's the case.

Thanks for contributing 🎉!

### DIFF
--- a/recipes/use_cases/email_agent/functions_prompt.py
+++ b/recipes/use_cases/email_agent/functions_prompt.py
@@ -8,7 +8,7 @@ list_emails_function = """
             "type": "dic",
             "properties": [
                 {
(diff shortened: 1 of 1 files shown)


IgorKasianenko:
`maxResults` as part of JSON the property must be camel-cased.
According to JSON.org: Property names must be camel-cased, ascii strings.

snowsky:
good catch, then somewhere else needs an update, I don't have time to look into that but should be a quick one.

PR #843, opened 2025-01-15
Update README.md

oneofth3m:
The path have changed in repo. Updating the cd command to reflect the same.


IgorKasianenko (approved):
Thanks for the fix!

PR #861, opened 2025-01-21
Fix package naming

init27:
Rename of library and imports to llama_cookbook


mreso (approved):
LGTM, lets fix the links and then merge this

PR #866, opened 2025-01-24
add MIT license to root

IgorKasianenko:
# What does this PR do?

Fixes # https://github.com/meta-llama/llama-cookbook/issues/749


carljparker:
Makes sense that this repo would use the MIT license. That is the license that is used by `llama-stack` and `llama-stack-apps`.

varunfb (approved):

PR #888, opened 2025-02-23
Update tool calling nbs to 3.3

init27:
Fix 101 and 201


IgorKasianenko:
Let me check if all the documentation updates from https://github.com/meta-llama/llama-cookbook/pull/885 are in place

PR #892, opened 2025-03-04
update test for guard prefix

IgorKasianenko:
# What does this PR do?

Fixes tests after PR https://github.com/meta-llama/llama-cookbook/pull/890

## Before submitting

Thanks for contributing 🎉!


albertodepaola (approved):
Fixing quickly to avoid failing tests. We need to follow up on proper legacy model support, as the new categories do not match older Llama models

PR #943, opened 2025-05-09
fix failing github actions

IgorKasianenko:
# What does this PR do?

Fixes failing GitHub actions

Thanks for contributing 🎉!


.github/workflows/pytest_cpu_gha_runner.yaml
```diff
@@ -70,7 +70,5 @@ jobs:
         id: test_summary
         uses: test-summary/action@v2
         with:
-          paths: |
-            **/*.xml
-            !**/AndroidManifest.xml
+          paths: $GITHUB_WORKSPACE/result.xml
```
seyeong-han, line 73: In the current github action the result failed as well.
I think it's because some projects don't have `result.xml` to check.

```
Error: ENOENT: no such file or directory, open '$GITHUB_WORKSPACE/result.xml'
```

IgorKasianenko:
@seyeong-han @maximgroshev updated the test

PR #950, opened 2025-05-20
Fix/update samsum

### LINKED ISSUE #948: Dataset 'Samsung/samsum' doesn't exist on the Hub or cannot be accessed.
I tried to finetune llama 3.1 8 b using samsum dataset but it shows it doesn't exit. Is there any other dataset or any other way i can load a custom dataset to  finetune it .

Image

nabidam:
# What does this PR do?

Fixes #948

This PR changes `Samsung/samsum` dataset repo url with an existing clone of that dataset `knkarthick/samsum`.
Original dataset has been deleted from huggingface recently (for any unknown reason).

## Before submitting
- [x] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).
- [x] Did you read the [contributor guideline](https://github.com/meta-llama/llama-cookbook/blob/main/CONTRIBUTING.md),
      Pull Request section?
- [x] Was this discussed/approved via a Github issue? Please add a link
      to it if that's the case.
- [x] Did you write any new necessary tests?

Thanks for contributing 🎉!

### DIFF
--- a/src/llama_cookbook/datasets/samsum_dataset.py
+++ b/src/llama_cookbook/datasets/samsum_dataset.py
@@ -8,16 +8,15 @@ import datasets
 
 from unittest.mock import patch
 
-
 @patch('builtins.input', return_value="N")
 def load_samsum(split, _):
     try:
         ds = datasets.load_dataset("knkarthick/samsum", split=split)
     except ValueError as e:
         if "trust_remote_code" in str(e):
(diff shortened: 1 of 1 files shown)


IgorKasianenko:
Hi, thanks for the PR! As the origin of the issue is coming from the HuggingFace, I will give it a day to see their response and will approve this if there is no better solution.

albertodepaola:
We should think about using datasets in a HF org we control meta-llama, to avoid this issue in the future. We can temporarily accept this while we work on a long term solution.

IgorKasianenko (approved):
LG

PR #951, opened 2025-05-20
Fix model string name for Llama-Prompt-Guard-2-86M in inference.py

carlini:
getting-started/responsible_ai/prompt_guard/inference.py has the incorrect model string for Prompt Guard 2. It's correct here

https://github.com/meta-llama/llama-cookbook/blob/7dab6a3298ec2e4742283670b56b6d405bafb54d/getting-started/responsible_ai/prompt_guard/inference.py#L22

but wrong here

https://github.com/meta-llama/llama-cookbook/blob/7dab6a3298ec2e4742283670b56b6d405bafb54d/getting-started/responsible_ai/prompt_guard/inference.py#L26


IgorKasianenko (approved):
Thanks for the fix!

PR #976, opened 2025-07-23
Create readme to point to new dirs

cmodi-meta:
# What does this PR do?

This PRs will create the old`/recipes` to only add readme files pointing to the new location of the recipes and getting-started guides.

This is to ensure that any existing links will not be broken and can point the user to the correct location.


cmodi-meta:
I've updated the PR to have a single old for NotebookLlama and make slight modification to the READMEs based on feedback.

connortreacy (approved):

PR #992, opened 2025-09-05
Delete package-lock.json to resolve Dependabot alerts #94 and #93

IgorKasianenko:
# What does this PR do?

Fixes https://github.com/meta-llama/llama-cookbook/security/dependabot/94 and https://github.com/meta-llama/llama-cookbook/security/dependabot/93

Thanks for contributing 🎉!


seyeong-han (approved):
It's okay to remove this `package-lock.json` file.
Doesn't have dependencies with the project.

PR #1013, opened 2025-10-31
Add code for measuring skills transferability across LoRA and FFT finetunes

subramen:
Add experiment code for:

- finetuning LoRA and FFT variants
- evaluating against 2 tasks
- measuring and comparing skills transfer from task1 to task2 across each LoRA/FFT variant


albertodepaola (approved):
Looks good overall! Would be good to add the package requirements in a single list for easier installment.

albertodepaola (approved):
Looks Good! Would be good to have a list of dependencies to install as well. 

As a second step, would also be good to see some of the example results and how to evaluate them

PR #1057, opened 2026-05-17
fix: guard choices access before indexing in LLM API calls (10 files)

qizwiz:
## Summary

LLM APIs (OpenAI, Groq, and OpenAI-compatible endpoints) can return an empty `choices` list in several legitimate scenarios:

- Content filtering triggered (safety system returns no completions)
- Rate-limit or quota errors during streaming
- Network-level partial responses
- Certain model configurations with `n=0`

Accessing `response.choices[0]` without a length check raises `IndexError` in these cases, turning a recoverable API error into an unhandled crash.

This PR adds an explicit `if not <response>.choices` guard before every unguarded `choices[0]` access across 10 files:

| File | Line | Fix |
|------|------|-----|
| `3p-integrations/groq/.../groq-quickstart-conversational-chatbot/main.py` | 35 | `continue` loop iteration |
| `3p-integrations/groq/.../presidential-speeches-rag-with-pinecone/main.py` | 65 | `return ""` |
| `3p-integrations/groq/.../text-to-sql-json-mode/main.py` | 33 | `return ""` |
| `3p-integrations/groq/.../verified-sql-function-calling/main.py` | 135 | `return None` |
| `end-to-end-use-cases/Contextual-Chunking-RAG/helper.py` | 30, 40 | `return ""` |
| `end-to-end-use-cases/Multi-Modal-RAG/scripts/final_demo.py` | 115, 184 | `return` fallback |
| `end-to-end-use-cases/github_triage/llm.py` | 39 | inline ternary guard |
| `end-to-end-use-cases/research_paper_analyzer/research_analyzer.py` | 98 | `return []` |
| `end-to-end-use-cases/structured_parser/src/utils.py` | 456 | `raise ValueError` |
| `getting-started/inference/api_inference/api_inference.py` | 40 | `return ""` |

Found by [`pact`](https://github.com/qizwiz/pact) static analysis (`llm_response_unguarded` mode), which uses a Z3 Fixedpoint Datalog engine to track unguarded LLM API response accesses across Python call graphs.

## Test plan

- [ ] Verify pact reports 0 `llm_response_unguarded` violations on patched files
- [ ] Run existing example tests if any exist in CI


qizwiz:
**Update: guard strengthened after live reproduction (commit 7e52db5)**

While building a proof script, we found the original guard (`if not response.choices`) is incomplete for one additional failure mode.

**Gemini 2.5 Flash** (OpenAI-compatible endpoint) returns HTTP 200 with `choices[0].message = None` when its safety filter blocks a request — `finish_reason: 'content_filter: PROHIBITED_CONTENT'`. The choices list has one entry so the length guard passes, but accessing `.message.content` raises `AttributeError: 'NoneType' object has no attribute 'content'`.

```python
# previous guard — still crashes on Gemini content filtering
if not response.choices:
    return ""
return response.choices[0].message.content  # AttributeError if message is None

# comprehensive fix (this commit)  
if not response.choices or response.choices[0].message is None:
    return ""
return response.choices[0].message.content  # safe
```

All 10 files updated. This covers: empty choices list (any provider), null message on content filter (Gemini and potentially others).

Note: still waiting on CLA confirmation before this can be reviewed — will sign at code.facebook.com/cla.

qizwiz:
@meta-cla-bot recheck

